PR #8, opened 2022-07-04
Add missing input bindings to DX11.

etra0:


veeenu:
LGTM, but I want to run some tests and try to understand a few things before merging it.

It's not clear to me why the imgui-rs `Key` enum doesn't have all the alphanumeric characters and that's part of the reason I have never implemented this in the first place, going for an external input control system in my implementors (i.e. via `GetAsyncKeyState` in ER and DS3 tools).

Furthermore, we're still on `imgui-rs` 0.8.0 by choice as the newer versions have broken docs and also, iirc, have some breaking changes despite only being patch version bumps.

I'd like to make sure that that's heading in the right direction, or alternatively sidestep that and rely on `imgui-sys` directly for this specific part.

etra0:
I was wondering that issue myself too. I saw this issue https://github.com/imgui-rs/imgui-rs/issues/455 which it seems to explain that those specified keys are for imgui-rs to handle directly.

For what its worth it, I checked some imgui backends (like the [SDL](https://github.com/imgui-rs/imgui-rs/blob/1de842a402af309adb1332c2b452ba7857fdc5b1/imgui-sdl2-support/src/lib.rs#L129-L150) and the [WINIT](https://github.com/imgui-rs/imgui-rs/blob/1de842a402af309adb1332c2b452ba7857fdc5b1/imgui-winit-support/src/lib.rs#L215-L236) and they only init those same keys in the same way. Without this patch, input fields won't capture those special keys (for example backspace, arrow keys, etc). The other inputs work fine.

It sounds more like a limitation if imgui-rs itself rather than imgui

veeenu:
Oh, yeah, indeed it is probably something specific to `imgui-rs`. That's why I suggested we find out a way to initialize all the keys in the appropriate way by peeling off a layer and using `imgui-sys` directly -- even simply checking out how the underlying index values are laid out, building our own `#[repr(usize)]` enum and using that for the indices.

I feel like I don't know that part of `imgui` well enough yet, but I don't think it'll be super hard to figure out.

veeenu (approved):

veeenu:
Turns out I'd already done the exact same thing for DX12 [here](
https://github.com/veeenu/hudhook/blob/master/hudhook/src/hooks/dx12.rs#L375-L401) 😅 I need to do some refactoring.

etra0:
Yeah, I noticed that as well but forgot to point it out haha. I can see a bit of code duplication that maybe could be reduced but I'm not sure if it would be _prettier_.

veeenu:
I think the style of the two hooks could be at least made more similar, but there are some very fundamental differences that completely prevent factoring common code together in a meaningful way.

Issue #10, opened 2022-07-05
WND proc not reset on DX12 hook


OneXDeveloper:
Hi,

Firstly, thanks for this project. 

I have been working on getting unhooking working on my tool and have got it working with one exception. The WND proc is not resetting after disabling hooks so the game will crash.

I didn’t want to PR my (probably wrong) approach because it felt pretty nasty and hacky. But the gist is that you need to call `SetWindowLongPtrA` with the original WND proc. You can get the original one by calling `GetWindowLongPtrA` with the hwnd in the `new` function of the `ImguiRenderer` and just save it somewhere until it’s time to shutdown. I had to create some functions to expose a shutdown method on the imgui renderer in the hook module and that felt weird. I’m sure you’d find a better way to handle that. 

Thanks again.

veeenu:
Oh, that is not a hack! There may be more or less clean ways of going about it, but definitely having a destructor kind of thing is a good idea -- after all, RAII is **the** pattern in Rust.

Feel free to share your PR if you'd like to! Even if it didn't fit with the code style for the rest of the code, it'd still be a great contribution and would help me figure out how to get some things right in the implementation -- and I could always request changes or modify it afterwards anyway.

Unhooking has been planned for a while, just never really figured out how to get it going properly. So thank you!

veeenu:
I noticed I am already storing the window proc in `ImguiRenderer`, see [this](https://github.com/veeenu/hudhook/blob/master/hudhook/src/hooks/dx12.rs#L523-L526); `cleanup` is invoked in the `ResizeBuffers` hook, like [this](
https://github.com/veeenu/hudhook/blob/master/hudhook/src/hooks/dx12.rs#L143-L162): the `ImguiRenderer` is just taken out of the `OnceCell` and later on recreated.

Is this similar to your solution?

OneXDeveloper:
The one you’re storing is the overridden one though. Not the original from the process. I wasn’t sure why you were doing that. But yes, very similar. The only difference really is I created a shutdown function, that I call from my detach, that did the same thing but with the one originally from the process. I obtained that with the get function I shared above. The unhooking and everything else was working great.

veeenu:
How/where did you write your detach function? I tried calling `FreeLibraryAndExitThread` from inside the UI loop but that resulted in crashing, as the function does not return and so its whole call stack doesn't get to do any cleanup.

> The one you’re storing is the overridden one though.

I'm not sure that's the case:

[...]

PR #13, opened 2022-07-07
Reduce code redundancy

etra0:
This is an attempt to reduce a bit code redundancy.

Although this is a bit unnecessary, I've already noticed a few discrepancies between the DX11 and DX12 implementations that actually do the same thing, like DX11 using some functions that DX12 doesn't, and the IO was un-synchronized for DX11 before my PR.

This is more a proposal rather than a fix. With this format, it's more harder to have different WndProc for both implementations so they can be better in sync.

I'm open to discuss this if it's something you'd be up to improve further on or simply drop it.

The `imgui_wnd_proc` is statically dispatched so it should have virtually no performance impact.

### DIFF
--- a/hudhook/src/hooks/common.rs
+++ b/hudhook/src/hooks/common.rs
@@ -11,8 +11,8 @@ pub(crate) type WndProcType =
 
 pub(crate) trait ImguiRendererInterface {
     fn io_mut(&mut self) -> &mut imgui::Io;
-    fn set_focus(&mut self, focus: bool);
-    fn is_focus(&self) -> bool;
+    fn get_focus_mut(&mut self) -> &mut bool;
+    fn get_focus(&self) -> bool;
     fn get_wnd_proc(&self) -> WndProcType;
 
     fn setup_io(&mut self) {
@@ -105,9 +105,9 @@ pub(crate) fn imgui_wnd_proc_impl(
         WM_CHAR => io.add_input_character(wparam as u8 as char),
         WM_ACTIVATE => {
             if loword(wparam as _) == WA_INACTIVE as u16 {
-                imgui_renderer.set_focus(false);
+                *imgui_renderer.get_focus_mut() = false;
             } else {
-                imgui_renderer.set_focus(true);
+                *imgui_renderer.get_focus_mut() = true;
             }
             return LRESULT(1);
         },
--- a/hudhook/src/hooks/common.rs
+++ b/hudhook/src/hooks/common.rs
@@ -1,0 +1,117 @@
+use imgui::*;
+use windows::Win32::Foundation::{HWND, LPARAM, LRESULT, WPARAM};
+use windows::Win32::UI::Input::KeyboardAndMouse::*;
+use windows::Win32::UI::WindowsAndMessaging::{WHEEL_DELTA, WM_XBUTTONDBLCLK, XBUTTON1, *};
+
+pub type WndProcType =
+    unsafe extern "system" fn(hwnd: HWND, umsg: u32, wparam: WPARAM, lparam: LPARAM) -> LRESULT;
+
+pub trait ImguiRendererCommon {
+    fn io_mut(&mut self) -> &mut imgui::Io;
+    fn set_focus(&mut self, focus: bool);
+    fn is_focus(&self) -> bool;
+    fn get_wnd_proc(&self) -> WndProcType;
+
+    fn init_io(&mut self) {
+        let mut io = ImguiRendererCommon::io_mut(self);
+
+        io.nav_active = true;
+        io.nav_visible = true;
+
+        // Initialize keys
+        io[Key::Tab] = VK_TAB.0 as _;
+        io[Key::LeftArrow] = VK_LEFT.0 as _;
+        io[Key::RightArrow] = VK_RIGHT.0 as _;
+        io[Key::UpArrow] = VK_UP.0 as _;
+        io[Key::DownArrow] = VK_DOWN.0 as _;
+        io[Key::PageUp] = VK_PRIOR.0 as _;
+        io[Key::PageDown] = VK_NEXT.0 as _;
+        io[Key::Home] = VK_HOME.0 as _;
+        io[Key::End] = VK_END.0 as _;
+        io[Key::Insert] = VK_INSERT.0 as _;
+        io[Key::Delete] = VK_DELETE.0 as _;
+        io[Key::Backspace] = VK_BACK.0 as _;
+        io[Key::Space] = VK_SPACE.0 as _;
+        io[Key::Enter] = VK_RETURN.0 as _;
+        io[Key::Escape] = VK_ESCAPE.0 as _;
+        io[Key::A] = VK_A.0 as _;
+        io[Key::C] = VK_C.0 as _;
+        io[Key::V] = VK_V.0 as _;
+        io[Key::X] = VK_X.0 as _;
+        io[Key::Y] = VK_Y.0 as _;
+        io[Key::Z] = VK_Z.0 as _;
+    }
+}
+
+pub struct WndProcResult(pub bool, pub Option<LRESULT>);
+
+#[must_use]
+pub(crate) fn imgui_wnd_proc_impl(
+    _hwnd: HWND,
+    umsg: u32,
+    WPARAM(wparam): WPARAM,
+    LPARAM(_lparam): LPARAM,
+    imgui_renderer: &mut Box<impl ImguiRendererCommon>,
+) -> WndProcResult {
+    let mut io = imgui_renderer.io_mut();
+    match umsg {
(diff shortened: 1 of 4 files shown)


etra0:
Of course if this is of any interest of you, I'll rebase this PR to a single commit in case you agree with the changes.

hudhook/src/hooks/common.rs
```diff
@@ -0,0 +1,119 @@
+use imgui::*;
+use parking_lot::MutexGuard;
+use windows::Win32::Foundation::{HWND, LPARAM, LRESULT, WPARAM};
+use windows::Win32::UI::Input::KeyboardAndMouse::*;
+use windows::Win32::UI::WindowsAndMessaging::{WHEEL_DELTA, WM_XBUTTONDBLCLK, XBUTTON1, *};
+
+pub(crate) type WndProcType =
+    unsafe extern "system" fn(hwnd: HWND, umsg: u32, wparam: WPARAM, lparam: LPARAM) -> LRESULT;
+
+pub(crate) trait ImguiRendererCommon {
```
veeenu, line 10: I'd change the `Common` suffix, or the name of the trait altogether, as I feel the current one doesn't communicate its intentions well enough. This is a nitpick, though, I don't have good alternative ideas for the time being so we can postpone it, or even merge and change it in another PR.

hudhook/src/hooks/common.rs
```diff
@@ -0,0 +1,119 @@
...
+    fn io_mut(&mut self) -> &mut imgui::Io;
+    fn set_focus(&mut self, focus: bool);
+    fn is_focus(&self) -> bool;
+    fn get_wnd_proc(&self) -> WndProcType;
+
+    fn init_io(&mut self) {
```
veeenu, line 16:
```suggestion
    fn setup_io(&mut self) {
```
I'd rename this to `setup` for clarity, as it is actually the content which is initialized rather than the `Io` itself.

hudhook/src/hooks/common.rs
```diff
@@ -0,0 +1,119 @@
...
+            io.mouse_down[0] = true;
+        },
+        WM_RBUTTONDOWN | WM_RBUTTONDBLCLK => {
+            io.mouse_down[1] = true;
+        },
+        WM_MBUTTONDOWN | WM_MBUTTONDBLCLK => {
+            io.mouse_down[2] = true;
+        },
+        WM_XBUTTONDOWN | WM_XBUTTONDBLCLK => {
+            let btn = if super::hiword(wparam as _) == XBUTTON1.0 as u16 { 3 } else { 4 };
```
veeenu, line 77: As a general principle, I'd prefer having these functions imported rather than referring to them by the full path.

hudhook/src/hooks/dx11.rs
```diff
@@ -223,37 +143,14 @@ impl ImguiRenderer {
...
-        io[Key::Y] = VK_Y.0 as _;
-        io[Key::Z] = VK_Z.0 as _;
-
         let flags = ImguiRenderLoopFlags { focused: true };
 
         trace!("Renderer initialized");
-        ImguiRenderer { engine, render_loop, wnd_proc, flags, swap_chain }
+        let mut renderer = ImguiRenderer { engine, render_loop, wnd_proc, flags, swap_chain };
+
+        ImguiRendererCommon::init_io(&mut renderer);
```
veeenu, line 151: I like this. Invoking the trait method as an associated function makes things clear.

hudhook/src/hooks/dx11.rs
```diff
@@ -310,6 +207,24 @@ impl ImguiRenderer {
     }
 }
 
+impl crate::hooks::common::ImguiRendererCommon for ImguiRenderer {
```
veeenu, line 210:
```suggestion
impl ImguiRendererCommon for ImguiRenderer {
```
Also here I'd rather just import the trait at the top.

hudhook/src/hooks/dx11.rs
```diff
@@ -310,6 +207,24 @@ impl ImguiRenderer {
...
+
+    fn set_focus(&mut self, focus: bool) {
+        self.flags.focused = focus;
+    }
+
+    fn is_focus(&self) -> bool {
+        self.flags.focused
+    }
+
+    fn get_wnd_proc(&self) -> crate::hooks::common::WndProcType {
```
veeenu, line 223:
```suggestion
    fn get_wnd_proc(&self) -> WndProcType {
```

hudhook/src/hooks/common.rs
```diff
@@ -0,0 +1,119 @@
...
+use windows::Win32::UI::Input::KeyboardAndMouse::*;
+use windows::Win32::UI::WindowsAndMessaging::{WHEEL_DELTA, WM_XBUTTONDBLCLK, XBUTTON1, *};
+
+pub(crate) type WndProcType =
+    unsafe extern "system" fn(hwnd: HWND, umsg: u32, wparam: WPARAM, lparam: LPARAM) -> LRESULT;
+
+pub(crate) trait ImguiRendererCommon {
+    fn io_mut(&mut self) -> &mut imgui::Io;
+    fn set_focus(&mut self, focus: bool);
+    fn is_focus(&self) -> bool;
```
veeenu, line 13: I see the two implementations for this are basically identical. It'd be great if we could do away with these altogether and find a different semantics for `focus`, but it's ok to leave these in for the time being.

veeenu:
Again, thank you so much for your contribution, this is really appreciated!

etra0:
Thanks for the feedback! I'll apply those changes today in the afternoon and then mark it as a proper PR.

hudhook/src/hooks/common.rs
etra0: hmm, any suggestion? I'd guess `ImguiRendererInterface` could be more appropriated? still unsure how to name it hehe.

hudhook/src/hooks/common.rs
etra0: Hmm from what I saw, on the standard libraries they usually do `get_something` and `get_something_mut`, maybe I could change the names to that to be more compliant with the usual naming scheme? otherwise, I'm not sure how to proceed.

veeenu: I am guilty of not having followed this myself so far, but the [API guidelines](https://rust-lang.github.io/api-guidelines/checklist.html) suggest to use `name` to get an immutable reference (or a copy, for `Copy` types) and `name_mut` for mutable references. I like this pattern.

(OT: we're making strides here so at some point I want to put out a new release, I want to go over that checklist and refactor everything up to spec)

etra0:
I think I applied most of the suggestions, the only thing I'm still not sure about is the name of the trait, I changed it from `ImguiRendererCommon` to `ImguiRendererInterface`, I still feel we could have a better name (wow, how hard can be naming things!?), but I think it's better to what I had.

I'm changing this PR to ready for review. Note that this PR conflicts with https://github.com/veeenu/hudhook/pull/14 since it effectively removes the branching too.

I also have some sort of patch to add some trait control for that issue but I think that's out of the scope for this PR.

hudhook/src/hooks/common.rs
etra0: Maybe it would be good to open an issue with such checklist? I'm happy to contribute whenever I can, I've been having a blast using this library tbh :)

hudhook/src/hooks/common.rs
veeenu: I read a bit from [here](https://github.com/rust-lang/api-guidelines/discussions/28) and then I realized we may go with a more radical approach here. The role of this trait is probably not limited to factoring out common behavior for DX impls but more generally could be the entry point to `WndProc` and other events related considerations.

I'd go for something like `WindowsEventHandler` (though I don't super like this either so feel free to come up with other suggestions -- it's an internal so we're never too late to change it). This would also give us a good spot where to put event related behavior extensions in the future (like `focus`).

hudhook/src/hooks/common.rs
veeenu: Great idea! Thank you!

veeenu (approved):

veeenu:
> I think I applied most of the suggestions, the only thing I'm still not sure about is the name of the trait, I changed it from `ImguiRendererCommon` to `ImguiRendererInterface`, I still feel we could have a better name (wow, how hard can be naming things!?), but I think it's better to what I had.

It never gets easier 😁 it's ok, I was going to work on `hudhook` a bit this morning anyway so I'll approve and merge this and then change it myself later.

> Note that this PR conflicts with #14 since it effectively removes the branching too.
> 
> I also have some sort of patch to add some trait control for that issue but I think that's out of the scope for this PR.

That's fine, we can close that one and figure out a better solution. I think this _could_ break #10 for @OneXDeveloper 's use case but they can specify revision in `Cargo.toml` dependencies while we address that.

Issue #12, opened 2022-07-06
No way to capture mouse input


OneXDeveloper:
Another modification I had to make to use this lib was reimplementing `is_capturing` on the `ImguiRenderLoop` trait. I wanted mouse clicks and text input to not pass through to the game when the imgui box is focused by the user. 

My exact implementation:

Trait impl:
```rust
fn is_capturing(&self, io: &imgui_dx12::imgui::Io) -> bool {
    io.want_capture_mouse
}
```

Default trait impl
```rust
fn is_capturing(&self, _: &imgui_dx12::imgui::Io) -> bool {
    false
}
```

Then near the bottom of `imgui_wnd_proc`:

```rust
if let Some(render_loop) = IMGUI_RENDER_LOOP.get() {
     if render_loop.is_capturing(imgui_renderer.ctx.io()) {
          return LRESULT(1);
     }
 }
 ```

Again, I don’t think this is the perfect implementation but it seems to work OK for my use case. If this could be implemented I believe I won’t have to vendor this lib anymore and can just use the one from git :)

**Comment by veeenu** [comment]
Would this work (in place of [this](https://github.com/veeenu/hudhook/blob/master/hudhook/src/hooks/dx12.rs#L271-L276))?

```rust
let want_capture = io.want_capture_mouse || io.want_capture_keyboard;
let wnd_proc = imgui_renderer.wnd_proc;
drop(imgui_renderer);

if want_capture {
    trace!("Leaving WndProc via capturing");
    LRESULT(1)
} else {
    trace!("Leaving WndProc via CallWindowProcW");
    CallWindowProcW(Some(wnd_proc), hwnd, umsg, WPARAM(wparam), LPARAM(lparam))
}
```

This should capture inputs in the way intended by imgui, and does not need to rely on augmenting the trait. Unfortunately it doesn't work for me as the games I built this for (DS3 & Elden Ring) seemingly don't even go through WndProc for gathering their input, but it is also not a problem for those so I can include it without introducing regressions.

It's already in `master`.

OneXDeveloper:
Yes, I think that would solve my case. I was opting for the trait the method to make it a bit more flexible for other use cases but it may not be needed. Awesome! Thanks again

PR #14, opened 2022-07-08
Partially revert 21f043d

etra0:
This patch crashes The Witcher 3 due to the chain of wnd_proc being
broken, so messages don't arrive to the parent window.

I'd say to properly ignore incoming messages would be to override the `wnd_proc` of the game itself, and peek the message.

### DIFF
--- a/hudhook/src/hooks/dx12.rs
+++ b/hudhook/src/hooks/dx12.rs
@@ -270,17 +270,12 @@ unsafe extern "system" fn imgui_wnd_proc(
                 _ => {},
             }
 
-            let want_capture = io.want_capture_mouse || io.want_capture_keyboard;
             let wnd_proc = imgui_renderer.wnd_proc;
             drop(imgui_renderer);
 
-            if want_capture {
-                trace!("Leaving WndProc via capturing");
-                LRESULT(1)
-            } else {
-                trace!("Leaving WndProc via CallWindowProcW");
-                CallWindowProcW(Some(wnd_proc), hwnd, umsg, WPARAM(wparam), LPARAM(lparam))
-            }
+            trace!("Leaving WndProc");
+
+            CallWindowProcW(Some(wnd_proc), hwnd, umsg, WPARAM(wparam), LPARAM(lparam))
         },
         Some(None) => {
             debug!("Could not lock in WndProc");


veeenu:
This would also revert and reopen #12. I'm not married to the solution that's being reverted in this PR but we do need to find a way to have the same functionality.

Could you try whether changing the `LRESULT(1)` return with a call to `DefWindowProcW` works? I think this could sidestep inputs that are directed to the game while preserving the overall windowing functionality. Unfortunately I don't have a means to test this. It would be nice to have a directory of clients of `hudhook` so changes could be tested over as many games as possible.

Pinging @OneXDeveloper in case they have other ideas that could work here.

OneXDeveloper:
I haven’t had a chance to test out this original change yet. But I wonder if the trait approach would work? That would allow us to selectively enable this based off of the IO or other logic of our tools.

etra0:
Here:

https://github.com/FransBouma/InjectableGenericCameraSystem/blob/90d7e39b2e9ff7736951a9ff47ac889707dc1072/Cameras/DMC5/InjectableGenericCameraSystem/InputHooker.cpp#L152-L164

You can see he hooked into GetMessage and then depending on the variables, he sets the message to NULL to avoid being processed by the windows procedure.

If I understand correctly, that would be the equivalent of hooking into the original windows procedure (the one it's called at the end of the *injected* one) and then in there decide to ignore the messages, or hook into the windows functions like GetMessage/PeekMessage and decide it there.

Cutting the chain right away means messages won't be handled to the parent procedure.

veeenu:
> Here:
> 
> https://github.com/FransBouma/InjectableGenericCameraSystem/blob/90d7e39b2e9ff7736951a9ff47ac889707dc1072/Cameras/DMC5/InjectableGenericCameraSystem/InputHooker.cpp#L152-L164
> 
> You can see he hooked into GetMessage and then depending on the variables, he sets the message to NULL to avoid being processed by the windows procedure.
> 
> If I understand correctly, that would be the equivalent of hooking into the original windows procedure (the one it's called at the end of the _injected_ one) and then in there decide to ignore the messages, or hook into the windows functions like GetMessage/PeekMessage and decide it there.

Oh, I see. Detouring those functions as well seems a bit much, I'd rather keep the profile as light as possible on that front and only detour what we absolutely need to.

> Cutting the chain right away means messages won't be handled to the parent procedure.
> 
> Also, the Trait method could work too since it would be set by the user rather than the library itself, which can be avoided for programs with UB. Since we're adding a trait for commonalities between DX11 and DX12 in #13, we could do it in there, but I'm not totally convinced that's the way to go yet.

If we could defer control of this to the user, with sensible defaults, I think that would be the best way to go. We could merge #13 once it's ready and then figure out a good solution starting from there.

etra0:
Superseeded by https://github.com/veeenu/hudhook/pull/13

PR #18, opened 2022-07-10
Merge ImguiRenderLoop trait & add should_block_messages.

etra0:
An attempt to reduce a bit more code. 

This also is an attempt to add the option to block messages if the user wants to.

### DIFF
--- a/hudhook/src/hooks/common.rs
+++ b/hudhook/src/hooks/common.rs
@@ -1,10 +1,12 @@
-use imgui::Key;
+use imgui::{Key, Context, Ui};
 use parking_lot::MutexGuard;
 use windows::Win32::Foundation::{HWND, LPARAM, LRESULT, WPARAM};
 use windows::Win32::UI::Input::KeyboardAndMouse::*;
 use windows::Win32::UI::WindowsAndMessaging::{WHEEL_DELTA, WM_XBUTTONDBLCLK, XBUTTON1, *};
 
-use super::{get_wheel_delta_wparam, hiword, loword};
+use super::dx11::ImguiDX11Hooks;
+use super::dx12::ImguiDX12Hooks;
+use super::{get_wheel_delta_wparam, hiword, loword, Hooks};
 
 pub(crate) type WndProcType =
     unsafe extern "system" fn(hwnd: HWND, umsg: u32, wparam: WPARAM, lparam: LPARAM) -> LRESULT;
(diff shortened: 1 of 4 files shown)


hudhook/src/hooks/common.rs
```diff
@@ -56,6 +58,7 @@ pub(crate) fn imgui_wnd_proc_impl(
     WPARAM(wparam): WPARAM,
     LPARAM(lparam): LPARAM,
     mut imgui_renderer: MutexGuard<Box<impl ImguiWindowsEventHandler>>,
+    imgui_render_loop: &Box<dyn ImguiRenderLoop + Send + Sync>
```
veeenu, line 61: We have to drop the renderer so it's ok to use the specialized mutex guard type, but here I'd go with a more generic `T: AsRef<ImguiRenderLoop + Send + Sync>` or something equivalent.

hudhook/src/hooks/common.rs
```diff
@@ -116,5 +119,43 @@ pub(crate) fn imgui_wnd_proc_impl(
...
+    /// initialize your data.
+    fn initialize(&mut self, _ctx: &mut Context) {}
+    /// Called every frame. Use the provided `ui` object to build your UI.
+    fn render(&mut self, ui: &mut Ui, flags: &ImguiRenderLoopFlags);
+
+    // If this function returns true, the WndProc function will not call the procedure of the
+    // parent window.
+    fn should_block_messages(&self) -> bool { false }
+
+    fn into_hook(self, backend: HookBackend) -> Box<dyn Hooks>
```
veeenu, line 153: I'd prefer avoiding runtime evaluations when we could statically rely on the type system. I'd instead use a trait for each backend that has a constructor, so that the user could invoke `.into_hook::<ImguiDX12Hooks>()` or something similar.

veeenu (changes requested):
This is very very good. Thank you!

etra0:
I made a new attempt, this time using `AsRef` and static dispatch for into_hook.

hudhook/src/hooks/common.rs
```diff
@@ -116,5 +122,52 @@ pub(crate) fn imgui_wnd_proc_impl(
     let wnd_proc = imgui_renderer.wnd_proc();
     drop(imgui_renderer);
 
+    if imgui_render_loop.as_ref().should_block_messages() {
```
OneXDeveloper, line 125: I believe we will need to pass in the IO or UI here to be able to correctly determine if we should block the message. The use case is blocking it when the UI wants IO.

veeenu: We _could_ have `Io` here, but not `Ui` as that object only lives during the invocation of `Present`. I don't have a strong opinion here; if we can pass in `ctx.io()` (immutable reference preferably) that would probably be less state management work on the clients, but for use cases that aren't 100% relying on that, clients would have to manage the state anyway. I'd say let's give this a try and change the function to accept `&Io`.

veeenu:
Sorry, been a little busy these past few days, will get to review as soon as I can!

veeenu (approved):

PR #25, opened 2022-08-27
Added OpenGL3 support

Jakobzs:
Still very WIP but open for review now 😄


hudhook/src/hooks/dx9.rs
```diff
@@ -51,12 +51,19 @@ unsafe fn draw(this: &IDirect3DDevice9) {
...
+            let mut imgui_renderer = ImguiRenderer {
                 ctx: context,
                 renderer,
                 wnd_proc,
                 flags: ImguiRenderLoopFlags { focused: false },
-            }))
+            };
+
+            // Initialize window events on the imgui renderer
+            ImguiWindowsEventHandler::setup_io(&mut imgui_renderer);
```
veeenu, line 63: @FrankvdStam I haven't tested DX9 much; do you think this is appropriate here, in the same way as DX11/12? Looks good to me but not sure if you left it out on purpose.

hudhook/src/hooks/opengl3.rs
```diff
@@ -0,0 +1,285 @@
...
+    fn wnd_proc(&self) -> WndProcType {
+        self.wnd_proc
+    }
+}
+unsafe impl Send for ImguiRenderer {}
+unsafe impl Sync for ImguiRenderer {}
+
+/// Stores hook detours and implements the [`Hooks`] trait.
+pub struct OpenGL3Hooks {
+    #[allow(dead_code)]
```
veeenu, line 218: I'd rather remove this and have warnings if the `RawDetour` is actually unused -- which it shouldn't anyway since it's in `unhook()`.

hudhook/src/hooks/opengl3.rs
```diff
@@ -0,0 +1,285 @@
...
+
+    fn wnd_proc(&self) -> WndProcType {
+        self.wnd_proc
+    }
+}
+unsafe impl Send for ImguiRenderer {}
+unsafe impl Sync for ImguiRenderer {}
+
+/// Stores hook detours and implements the [`Hooks`] trait.
+pub struct OpenGL3Hooks {
```
veeenu, line 217: I'd rename this as `OpenGl3Hooks` to comply with the [API guidelines](https://rust-lang.github.io/api-guidelines/naming.html) that state:

> In UpperCamelCase, acronyms and contractions of compound words count as one word: use `Uuid` rather than `UUID`, `Usize` rather than `USize` or `Stdin` rather than `StdIn`.

Chances are this is a refactor that has to be done elsewhere in the codebase too, but it's worth it to do it as we go.

hudhook/src/hooks/opengl3.rs
```diff
@@ -0,0 +1,285 @@
...
+
+        if let Some(renderer) = IMGUI_RENDERER.take() {
+            renderer.lock().cleanup();
+        }
+        drop(IMGUI_RENDER_LOOP.take());
+    }
+}
+
+// Get the address of wglSwapBuffers in opengl32.dll
+unsafe fn get_opengl_wglswapbuffers_addr() -> OpenGl32wglSwapBuffers {
```
veeenu, line 273: Can we move this before the `OpenGL3Hooks` struct definition? I'd prefer having dependent functions below dependencies.

imgui-opengl3/src/lib.rs
```diff
@@ -0,0 +1,3 @@
+// TODO: Consider making local versions of gl_loader and imgui opengl renderer,
+// but this works for the time being
+pub use {gl_loader, imgui_opengl_renderer};
```
veeenu, line 3: This is perfectly ok if it works for current use cases, but yeah, I'd eventually like to embed our own implementations if possible.

veeenu (changes requested):
Thank you for this PR, there are just a few minor style concerns to address but it overall looks good to me!

hudhook/src/hooks/common.rs
```diff
@@ -7,6 +7,7 @@ use windows::Win32::UI::WindowsAndMessaging::{WHEEL_DELTA, WM_XBUTTONDBLCLK, XBU
 use super::dx11::ImguiDX11Hooks;
 use super::dx12::ImguiDX12Hooks;
 use super::dx9::ImguiDX9Hooks;
+use super::opengl3::OpenGL3Hooks;
```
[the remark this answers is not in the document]
Jakobzs: Renamed to ImguiOpenGl3Hooks (due to the comment below about renaming OpenGL3Hooks)

hudhook/src/hooks/dx9.rs
Jakobzs: Awaiting reply here is probably the best yep

hudhook/src/hooks/opengl3.rs
Jakobzs: Renamed

hudhook/src/hooks/opengl3.rs
Jakobzs: Done

imgui-opengl3/src/lib.rs
Jakobzs: Agreed :)

veeenu (approved):
Waiting on feedback wrt `setup_io` but otherwise LGTM!

hudhook/src/hooks/dx9.rs
FrankvdStam: Nice seeing an openGL target being added, nice work Jakobzs!
As for this line of code, I'm honestly not sure, I still didn't find the time to actually use the dx9 target in a real application. Did you actually try, and had problems, leading to you adding this line?

Jakobzs: Thank you :D

I have honestly not tested it on a DX9 target as I do not know one currently (all I know are DX11 and up). But from reasoning of that it was an issue in this hook (hence the added code), and also I based a lot of it on yours, then I assumed it might have been an issue there as well. It is a bit out of scope of this pull request tbh, but could be a nice fix if it is really needed.

FrankvdStam: If these are window events like resizing imgui windows, then that is already working without this.  There's at least 1 bug that I know of, which is that the game will crash when changing the resolution. Not sure if this could be related.

Maybe its better to just leave this alone until one of us actually uses it in a dx9 game, thats' when we'll find all the bugs.

Jakobzs: Could probably be ideal yeah. I will just revert it.

PR #59, opened 2022-12-01
Fix for drawing more than 64k draw commands

joffreybesos:
In ImGui there was once a limitation with more than 64k vertexes. To fix this you set this flag
`ctx.io_mut().backend_flags |= BackendFlags::RENDERER_HAS_VTX_OFFSET;`
Which was already done correctly in hudhook, however you need to then use the DrawCmd VtxOffset value when iterating through the draw commands.

For reference:
https://skia.googlesource.com/external/github.com/ocornut/imgui/+/refs/heads/features/draw_callback_extra_arg/examples/imgui_impl_dx12.cpp#231


veeenu (approved):
LGTM!

PR #64, opened 2022-12-18
feat: Set the special keys

etra0:
This allows to use things that require those modifiers, for example, the `Io::font_allow_user_scaling` requires for imgui to be able to detect the control key.

This fixes the behavior of CTRL+Click to manually input a value and SHIFT+drag to change range values more quickly.

The implementation is based on the winit impl and they do the double check just fine [1].

[1] https://github.com/imgui-rs/imgui-rs/blob/b1e66d050e84dbb2120001d16ce59d15ef6b5303/imgui-winit-support/src/lib.rs#L401-L404

### DIFF
--- a/src/hooks/common.rs
+++ b/src/hooks/common.rs
@@ -70,15 +70,22 @@ where
 {
     let mut io = imgui_renderer.io_mut();
     match umsg {
-        WM_KEYDOWN | WM_SYSKEYDOWN => {
-            if wparam < 256 {
-                io.keys_down[wparam] = true;
-            }
-        },
-        WM_KEYUP | WM_SYSKEYUP => {
-            if wparam < 256 {
-                io.keys_down[wparam] = false;
-            }
+        state @ (WM_KEYDOWN | WM_SYSKEYDOWN | WM_KEYUP | WM_SYSKEYUP) if wparam < 256 => {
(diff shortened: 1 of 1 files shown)


etra0:
woops, sorry for the force push, I initially did `<= 256`  when it should be strict less.

etra0:
Converted to draft as I missed some virtual keys and I'm away from my windows machine.

veeenu:
Looks good to me -- I wasn't aware of this shortcoming so it's all the more welcome!

etra0:
Now I added the rest of the special keys :D

etra0:
> Looks good to me -- I wasn't aware of this shortcoming so it's all the more welcome!

This is surprisingly useful for users with high resolution monitors! The use case of the litcher is to take photos inside games, so folks usually use 'hotsample' (i.e. run the game on a higher resolution than the monitor can support) to take a photo with amazing quality, but that means imgui font will be tiny, so someone suggested me to use the font scaling native from imgui and now with this it works!

You have to enable it manually, I did it on the [initialize](https://github.com/etra0/litcher/blob/46f31036713806c601f65966acb9f6d3d2349dc2/src/lib.rs#L269-L272) function, I thought about doing a PR for the default but wasn't sure if that's preferred.

veeenu (approved):

PR #65, opened 2022-12-28
Fixed OpenGL3 for imgui 0.9.0

Jakobzs:


veeenu (changes requested):
Looks great! Can you remove the `lib/minhook` submodule? We've already vendored it in `vendor/hudhook`. Once that's done, we can merge!

Issue #73, opened 2023-01-26
Initialize runs several times in a few games


Godnoken:
I noticed that the ImguiRenderLoop initialize function will run several times if the DLL is injected early in certain games.

Examples: Mount & Blade: Warband (32bit, DX9), Niche (64bit, DX11), Oxygen Not Included (64bit, DX11).

Is this something you're aware of @veeenu or possibly have an idea of why that is happening?

The side effect of this is that the game input gets all whacked up. Sometimes a keypress or mouseclick gets 'locked' to keydown and sometimes you have to click several times before anything registers.

Godnoken:
https://github.com/veeenu/hudhook/blob/4c2968d23d70d963e094db1ffacc8eeeba40c8f6/src/hooks/dx11.rs#L77

I know next to nothing about swap chains, but it seems that the ImguiRenderer gets invoked more than once inside of it in some cases.

veeenu:
Hi! Sorry for the delay, been quite busy.

In some games (definitely DS3 and ER) not reinitializing the renderer means that the UI won't scale alongside resizing the window, at best, or crash, at worst.

The fact that it gets called multiple times with early injections makes me think that resize is just called pointlessly by the game/engine a number of times while nothing is still getting rendered yet -- I wouldn't say this is something `hudhook` should work around.

Godnoken:
> Hi! Sorry for the delay, been quite busy.

No worries, no pressure, just happy to get a response eventually!

> In some games (definitely DS3 and ER) not reinitializing the renderer means that the UI won't scale alongside resizing the window, at best, or crash, at worst.

Aha! I suppose I should just make changes on my own end then. I have some functionality in the initialize function that I'd rather not have run more than once.

> The fact that it gets called multiple times with early injections makes me think that resize is just called pointlessly by the game/engine a number of times while nothing is still getting rendered yet -- I wouldn't say this is something `hudhook` should work around.

Absolutely correct, thank you for the input mate. I'll close this one.

Issue #72, opened 2023-01-16
DX9 Eject & Reset crash


AyserJamshidi:
When utilizing ImguiDx9Hooks, calling eject will cause a crash as the ImguiRenderer cleanup does nothing.  This can be slightly remedied by implementing it as:
```rust
unsafe fn cleanup(&mut self) {
    #[cfg(any(target_arch = "aarch64", target_arch = "x86_64"))]
    SetWindowLongPtrA(self.renderer.get_hwnd(), GWLP_WNDPROC, self.wnd_proc as usize as isize);

    #[cfg(target_arch = "x86")]
    SetWindowLongA(self.renderer.get_hwnd(), GWLP_WNDPROC, self.wnd_proc as usize as i32);
}
```

Note that even with this cleanup implementation **sometimes** ejecting (in a previously injected game process) will still cause a crash, but not every time.  This happens even when utilizing the hudhook! macro and using a modified version of the README ImGui window that contains only a button that calls eject.

This is not a whole solution as DX9 will still crash when/after Reset gets called.  I'm unsure if it's directly caused by Reset, but when resizing a game window (Aion MMO in this case), it will call Reset, which promptly crashes in the sequence of: 
* Cursor starts resizing Game window by dragging the window border
* Game turns to black due to resize (this is normal behavior for this game)
* Game renders again with the ImGui overlay for one frame
* Game freezes/crashes

I have not made a PR for this fix due to the Reset issue, but implementing cleanup and waiting a bit before calling eject does improve the DX9 experience.  

If any extra information is needed on these issues please let me know.  Thank you

veeenu:
Thank you for the feedback!

I'm assuming adding some sort of [synchronization](https://github.com/veeenu/hudhook/blob/main/src/hooks/dx12.rs#L800-L828) may be in order, to make sure we don't hard kill stuff that needs proper cleanup. That could be the culprit for the remainder of crashes, but resetting the correct window proc is already a good step towards that.

Feel free to submit a PR -- it's ok if issues get fixed gradually!

veeenu:
Closing this as most likely fixed by #74 -- we need more evidence to make sure, but we can reopen if the issue persists.

PR #74, opened 2023-02-03
Fix resizing crashes & potentially eject crashes

Godnoken:
This fixes crashes on resize/reset with DX9 games and likely most of the ejection crashes as discussed here https://github.com/veeenu/hudhook/issues/72

Funny enough - I had seen that Github issue before but didn't realize that was why I had issues with some DX9 games crashing on launch due to them calling the resize/reset function multiple times. Thanks to https://github.com/veeenu/hudhook/issues/73#issuecomment-1412476482 I understood where to look.

I have not yet tried ejecting but I assume that it will still crash sometimes as mentioned in the issue, unless this reset fix makes that go away completely.

### DIFF
--- a/src/hooks/dx9.rs
+++ b/src/hooks/dx9.rs
@@ -125,7 +125,7 @@ unsafe extern "system" fn imgui_wnd_proc(
     WPARAM(wparam): WPARAM,
     LPARAM(lparam): LPARAM,
 ) -> LRESULT {
-    if let Some(Some(imgui_renderer)) = IMGUI_RENDERER.get_mut().map(|m| m.try_lock()) {
+    if let Some(imgui_renderer) = IMGUI_RENDERER.get_mut().and_then(|m| m.try_lock()) {
         imgui_wnd_proc_impl(
             hwnd,
             umsg,
--- a/src/hooks/dx9.rs
+++ b/src/hooks/dx9.rs
@@ -125,21 +125,15 @@ unsafe extern "system" fn imgui_wnd_proc(
     WPARAM(wparam): WPARAM,
     LPARAM(lparam): LPARAM,
 ) -> LRESULT {
-    if IMGUI_RENDERER.get().is_some() {
-        match IMGUI_RENDERER.get_mut().unwrap().try_lock() {
-            Some(imgui_renderer) => imgui_wnd_proc_impl(
-                hwnd,
-                umsg,
-                WPARAM(wparam),
-                LPARAM(lparam),
-                imgui_renderer,
-                IMGUI_RENDER_LOOP.get().unwrap(),
-            ),
-            None => {
-                debug!("Could not lock in WndProc");
-                DefWindowProcW(hwnd, umsg, WPARAM(wparam), LPARAM(lparam))
-            },
-        }
+    if let Some(Some(imgui_renderer)) = IMGUI_RENDERER.get_mut().map(|m| m.try_lock()) {
+        imgui_wnd_proc_impl(
+            hwnd,
+            umsg,
+            WPARAM(wparam),
+            LPARAM(lparam),
+            imgui_renderer,
+            IMGUI_RENDER_LOOP.get().unwrap(),
+        )
     } else {
         debug!("WndProc called before hook was set");
         DefWindowProcW(hwnd, umsg, WPARAM(wparam), LPARAM(lparam))
--- a/src/hooks/dx9.rs
+++ b/src/hooks/dx9.rs
@@ -29,7 +29,7 @@ use crate::renderers::imgui_dx9;
 
 unsafe fn draw(this: &IDirect3DDevice9) {
     let mut imgui_renderer = IMGUI_RENDERER
-        .get_or_insert_with(|| {
+        .get_or_init(|| {
             let mut context = imgui::Context::create();
             context.set_ini_filename(None);
             IMGUI_RENDER_LOOP.get_mut().unwrap().initialize(&mut context);
@@ -87,7 +87,9 @@ unsafe extern "system" fn imgui_dx9_reset_impl(
         (*present_params).BackBufferHeight
     );
 
-    IMGUI_RENDERER = None;
+    if let Some(renderer) = IMGUI_RENDERER.take() {
+        renderer.lock().cleanup();
+    }
 
     let (_, _, trampoline_reset) =
         TRAMPOLINE.get().expect("IDirect3DDevice9::Reset trampoline uninitialized");
@@ -123,8 +125,8 @@ unsafe extern "system" fn imgui_wnd_proc(
     WPARAM(wparam): WPARAM,
     LPARAM(lparam): LPARAM,
 ) -> LRESULT {
-    if IMGUI_RENDERER.is_some() {
-        match IMGUI_RENDERER.as_mut().unwrap().try_lock() {
+    if IMGUI_RENDERER.get().is_some() {
+        match IMGUI_RENDERER.get_mut().unwrap().try_lock() {
             Some(imgui_renderer) => imgui_wnd_proc_impl(
                 hwnd,
                 umsg,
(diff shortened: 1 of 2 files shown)


src/hooks/dx9.rs
```diff
@@ -123,8 +125,8 @@ unsafe extern "system" fn imgui_wnd_proc(
     WPARAM(wparam): WPARAM,
     LPARAM(lparam): LPARAM,
 ) -> LRESULT {
-    if IMGUI_RENDERER.is_some() {
-        match IMGUI_RENDERER.as_mut().unwrap().try_lock() {
+    if IMGUI_RENDERER.get().is_some() {
+        match IMGUI_RENDERER.get_mut().unwrap().try_lock() {
```
veeenu, line 129: Small nit, just noticed this now; even though it's not pertinent to this PR, while we're at it I'd lighten up the syntax by compressing the if and match in the same statement:
```suggestion
        if let Some(Ok(renderer)) = IMGUI_RENDERER.get_mut().map(|m| m.try_lock()) {
```

veeenu (changes requested):
Thank you for your PR, looks pretty neat!

Can you try ejecting and run a few more manual tests? Once that's done, I'd say lgtm.

src/hooks/dx9.rs
Godnoken: Done. However, I had to use Some(Some()) instead, a bit ugly but eh, it was ugly anyways 😄

Godnoken:
> Thank you for your PR, looks pretty neat!

Cheers! My first real contribution to OSS.

> Can you try ejecting and run a few more manual tests? Once that's done, I'd say lgtm.

Done & done.

I've had no issues with resize crashes so far. Ejecting does unfortunately still crash sometimes. Some games work flawlessly while other doesn't seem to like it all the time.

I tried;

- M&B: Warband
- W40K
- Shelter 2
- Fallout 3
- Killing Floor
- Tropico 4
- Victoria II
-

src/hooks/dx9.rs
veeenu: Oh, in that case, my bad, thought it was an `Option<Result>`. In those cases, `and_then` in place of `map` has you covered 😅

veeenu:
> > Thank you for your PR, looks pretty neat!
> 
> Cheers! My first real contribution to OSS.

💪 

> I've had no issues with resize crashes so far. Ejecting does unfortunately still crash sometimes. Some games work flawlessly while other doesn't seem to like it all the time.

No worries, I think it's heavily dependent on what the game does anyway, and besides even DX11&12 are affected by the same instability.
 
> Do you think it is a possibility that SetWindowLongA/PtrA is calling it on the wrong window in some games? DX11 and DX12 uses the swapchain to get the correct window, while DX9 uses [D3DDEVICE_CREATION_PARAMETERS](https://learn.microsoft.com/en-us/windows/win32/direct3d9/d3ddevice-creation-parameters). I'm thinking that hFocusWindow isn't guaranteed to be the correct window if the game uses multiple.

I think most of our methods are heuristics anyway -- if we find a better one than `hFocusWindow`, then I'm all for switching over. It would probably be worth even if it was a more expensive method, as it's going to be run only once anyway (or once per resize, at worst).

> I just noticed the OpenGL3 hook doesn't have a proper cleanup fn either, I'll get that fixed too.

Great, thank you 🙏

Godnoken:
> No worries, I think it's heavily dependent on what the game does anyway, and besides even DX11&12 are affected by the same instability.

Oh really? Hmm.. 

> I think most of our methods are heuristics anyway -- if we find a better one than `hFocusWindow`, then I'm all for switching over. It would probably be worth even if it was a more expensive method, as it's going to be run only once anyway (or once per resize, at worst).

I'm trying it out on OpenGL3 now & it also crashes sometimes. The OpenGL implementation uses [WindowFromDC](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-windowfromdc), so in that case - that's a third option that doesn't work.

I'll see if I can catch the crash error & investigate further.

Btw - [SetWindowLongPtr](https://learn.microsoft.com/en-us/windows/win32/api/winuser/nf-winuser-setwindowlongptra) is the only one needed for both 32-bit and 64-bit according to its description. Shall I drop SetWindowLong from all hooks?

src/hooks/dx9.rs
Godnoken: I'm not too sure of how to write it as my idiomatic Rust skills are not quite there yet.. 😄 

I'll leave it as is for now.

Godnoken:
I've never used debuggers in games like this so I honestly am not too sure of how to use it.. `User-mode data execution prevention (DEP) violation at location` is the error I got from lldb when crashing. Doesn't tell me much. Eitherway, DX9 seems to crash either before or during `FreeLibraryAndExitThread` while OpenGL seemingly only crashes during `FreeLibraryAndExitThread`.

I feel like this is out of my depth and I have more pressing things to take care of at the moment.

Think we're good to merge this for now?

src/hooks/opengl3.rs
```diff
@@ -184,7 +184,13 @@ impl ImguiRenderer {
         self.renderer.render(&mut self.ctx);
     }
 
-    unsafe fn cleanup(&mut self) {}
+    unsafe fn cleanup(&mut self) {
+        #[cfg(any(target_arch = "aarch64", target_arch = "x86_64"))]
+        dbg!(SetWindowLongPtrA(self.game_hwnd, GWLP_WNDPROC, self.wnd_proc as usize as isize));
+
+        #[cfg(target_arch = "x86")]
+        dbg!(SetWindowLongA(self.game_hwnd, GWLP_WNDPROC, self.wnd_proc as usize as i32));
```
veeenu, line 192: Could you please remove the `dbg!` here?

I think we _should_ probably manage the resulting errors but we should emit a well formed log entry instead. We can do it in the future, in a dedicated refactoring PR.

src/hooks/dx9.rs
```diff
@@ -123,21 +125,15 @@ unsafe extern "system" fn imgui_wnd_proc(
...
-                LPARAM(lparam),
-                imgui_renderer,
-                IMGUI_RENDER_LOOP.get().unwrap(),
-            ),
-            None => {
-                debug!("Could not lock in WndProc");
-                DefWindowProcW(hwnd, umsg, WPARAM(wparam), LPARAM(lparam))
-            },
-        }
+    if let Some(Some(imgui_renderer)) = IMGUI_RENDERER.get_mut().map(|m| m.try_lock()) {
```
veeenu, line 128:
```suggestion
    if let Some(imgui_renderer) = IMGUI_RENDERER.get_mut().and_then(|m| m.try_lock()) {
```

veeenu (changes requested):
Couple minor things and then we're good to go.

veeenu:
> Scratch that, 32-bit implementation of SetWindowLongPtr doesn't exist in the windows crate until 0.41.0 and we're one 0.39.0. Any plans to bump up?

Absolutely. Haven't been able to keep tabs on `windows-rs` but I'll try and make the bump PR as soon as I get the chance.

> I feel like this is out of my depth and I have more pressing things to take care of at the moment.
> 
> Think we're good to merge this for now?

Regardless of that, I feel like this PR is very worthwhile as the semantics around `OnceCell` are better than `Option`.
So, resize crashes are gone and eject crashes are occasional still? If that's the case, it's still a huge improvement.

src/hooks/opengl3.rs
Godnoken: Woops! Didn't mean to leave it like that, my bad 😆

Godnoken:
> Absolutely. Haven't been able to keep tabs on `windows-rs` but I'll try and make the bump PR as soon as I get the chance.

Excellent!

> Regardless of that, I feel like this PR is very worthwhile as the semantics around `OnceCell` are better than `Option`. So, resize crashes are gone and eject crashes are occasional still? If that's the case, it's still a huge improvement.

I have personally not experienced any resize crashes but eject crashes do indeed still happen sometimes.

There's probably a better way to do the resizing too as it is not as smooth as in the other renderers - but it works, and how often do you resize really?

veeenu:
> There's probably a better way to do the resizing too as it is not as smooth as in the other renderers - but it works, and how often do you resize really?

It is something that I expect regular users only do when they want to switch their game resolution up or down every once in a while; I see most games nowadays only support a few different somewhat standard resolutions and handle all of that themselves anyway, so yeah, not terribly critical. It's just important that it works correctly but it doesn't need to be mega fast.

veeenu (approved):
LGTM!

PR #75, opened 2023-02-05
Add internal OpenGL loader

Godnoken:
Instead of an external dependency, I figured I'd write my own OpenGL loader.

I wrote it a few weeks ago and it's not beautiful, but it has worked fine for me, although it doesn't have Linux or Mac support (..nor does this crate, right..? 😄 )

### DIFF
--- a/src/renderers/imgui_opengl3.rs
+++ b/src/renderers/imgui_opengl3.rs
@@ -1,26 +1,32 @@
 use std::ffi::{c_void, CString, OsStr};
 use std::os::windows::prelude::OsStrExt;
 
+use once_cell::sync::OnceCell;
+use parking_lot::Mutex;
 use windows::core::{PCSTR, PCWSTR};
-use windows::Win32::Foundation::{FARPROC, PROC};
+use windows::Win32::Foundation::{FARPROC, HINSTANCE, PROC};
 use windows::Win32::Graphics::OpenGL::wglGetProcAddress;
 use windows::Win32::System::LibraryLoader::{GetProcAddress, LoadLibraryW};
 
-pub unsafe fn get_proc_address(function_string: &str) -> *const c_void {
+static mut OPENGL3_LIB: OnceCell<Mutex<HINSTANCE>> = OnceCell::new();
+
+unsafe fn get_opengl3_lib() -> HINSTANCE {
     let mut opengl_wide_string: Vec<u16> = OsStr::new("opengl32.dll").encode_wide().collect();
     opengl_wide_string.push(0);
 
-    let module = LoadLibraryW(PCWSTR(opengl_wide_string.as_ptr())).unwrap();
+    LoadLibraryW(PCWSTR(opengl_wide_string.as_ptr())).unwrap()
+}
+
+pub unsafe fn get_proc_address(function_string: &str) -> *const c_void {
+    let module = OPENGL3_LIB.get_or_init(|| Mutex::new(get_opengl3_lib())).lock();
 
     let function_c_string = CString::new(function_string).unwrap();
-    let function_bytes_with_nul: &[u8] = function_c_string.as_bytes_with_nul();
-
-    let function_name_ptr: *const u8 = function_bytes_with_nul.as_ptr() as *const u8;
+    let function_name_ptr = function_c_string.as_bytes_with_nul().as_ptr();
 
     let wgl_proc_address: PROC = wglGetProcAddress(PCSTR(function_name_ptr));
 
     if wgl_proc_address.is_none() {
-        let proc_address: FARPROC = GetProcAddress(module, PCSTR(function_name_ptr));
+        let proc_address: FARPROC = GetProcAddress(*module, PCSTR(function_name_ptr));
         proc_address.unwrap() as _
     } else {
         wgl_proc_address.unwrap() as _
--- a/src/renderers/imgui_opengl3.rs
+++ b/src/renderers/imgui_opengl3.rs
@@ -1,3 +1,28 @@
-// TODO: Consider making local versions of gl_loader and imgui opengl renderer,
-// but this works for the time being
-pub use {gl_loader, imgui_opengl_renderer};
+use std::ffi::{c_void, CString, OsStr};
+use std::os::windows::prelude::OsStrExt;
+
+use windows::core::{PCSTR, PCWSTR};
+use windows::Win32::Foundation::{FARPROC, PROC};
+use windows::Win32::Graphics::OpenGL::wglGetProcAddress;
+use windows::Win32::System::LibraryLoader::{GetProcAddress, LoadLibraryW};
+
+pub unsafe fn get_proc_address(function_string: &str) -> *const c_void {
+    let mut opengl_wide_string: Vec<u16> = OsStr::new("opengl32.dll").encode_wide().collect();
+    opengl_wide_string.push(0);
+
+    let module = LoadLibraryW(PCWSTR(opengl_wide_string.as_ptr())).unwrap();
+
+    let function_c_string = CString::new(function_string).unwrap();
+    let function_bytes_with_nul: &[u8] = function_c_string.as_bytes_with_nul();
+
+    let function_name_ptr: *const u8 = function_bytes_with_nul.as_ptr() as *const u8;
+
+    let wgl_proc_address: PROC = wglGetProcAddress(PCSTR(function_name_ptr));
+
+    if wgl_proc_address.is_none() {
+        let proc_address: FARPROC = GetProcAddress(module, PCSTR(function_name_ptr));
+        proc_address.unwrap() as _
+    } else {
+        wgl_proc_address.unwrap() as _
+    }
+}
--- a/src/renderers/imgui_opengl3.rs
+++ b/src/renderers/imgui_opengl3.rs
@@ -1,13 +1,12 @@
 use std::ffi::{c_void, CString};
 
 use once_cell::sync::OnceCell;
-use parking_lot::Mutex;
 use windows::core::PCSTR;
 use windows::Win32::Foundation::{FARPROC, HINSTANCE};
 use windows::Win32::Graphics::OpenGL::wglGetProcAddress;
 use windows::Win32::System::LibraryLoader::{GetProcAddress, LoadLibraryA};
 
-static mut OPENGL3_LIB: OnceCell<Mutex<HINSTANCE>> = OnceCell::new();
+static mut OPENGL3_LIB: OnceCell<HINSTANCE> = OnceCell::new();
 
 unsafe fn get_opengl3_lib() -> HINSTANCE {
     let opengl3_cstring = CString::new("opengl32.dll").unwrap();
@@ -19,7 +18,7 @@ unsafe fn get_opengl3_lib() -> HINSTANCE {
 ///
 /// Help me out lol
 pub unsafe fn get_proc_address(function_string: CString) -> *const c_void {
-    let module = OPENGL3_LIB.get_or_init(|| Mutex::new(get_opengl3_lib())).lock();
+    let module = OPENGL3_LIB.get_or_init(|| get_opengl3_lib());
 
     if let Some(wgl_proc_address) = wglGetProcAddress(PCSTR(function_string.as_ptr() as _)) {
         wgl_proc_address as _
--- a/src/renderers/imgui_opengl3.rs
+++ b/src/renderers/imgui_opengl3.rs
@@ -6,7 +6,7 @@ use windows::Win32::Foundation::{FARPROC, HINSTANCE};
 use windows::Win32::Graphics::OpenGL::wglGetProcAddress;
 use windows::Win32::System::LibraryLoader::{GetProcAddress, LoadLibraryA};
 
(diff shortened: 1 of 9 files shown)


veeenu:
Thank you for your PR! Glad to have this. In general, the fewer external dependencies, the better it is for this project I feel.

It looks good to me, pinging @Jakobzs for a review as they've worked on most of OpenGL3 related stuff and may have some suggestions!

To answer your doubt, yeah, we're only supporting Windows as a use case. It's a bit unfortunate but this is very hard to make cross platform -- and it's already skirting on shaky assumptions of behavior which is officially unsupported. 😅

Godnoken:
> Thank you for your PR! Glad to have this. In general, the fewer external dependencies, the better it is for this project I feel.

Good to hear, I'm really enjoying working on these things as I feel like they are at a good level of difficulty & complexity for me to develop.

> It looks good to me, pinging @Jakobzs for a review as they've worked on most of OpenGL3 related stuff and may have some suggestions!

Superb!

> To answer your doubt, yeah, we're only supporting Windows as a use case. It's a bit unfortunate but this is very hard to make cross platform -- and it's already skirting on shaky assumptions of behavior which is officially unsupported. 😅

Thought so. I'm not a big fan of other OSs but Linux isn't out of the question for me in the future as a user..

Godnoken:
> It's a first for me, too -- it's my first project gathering any attention to it at all, and requiring some maintenance!

Well deserved too! It's been an essential crate for me so far & I'm learning a lot from it.

> I heard Windows emulation support on Linux is pretty good nowadays -- strides were made in cross platform support with Proton considering that the Steam Deck runs Linux itself, so that might not be a huge problem for that use case at least. For Linux-native games, that won't cut it though.

Oh yeah, forgot about the whole Steam Deck thingy. Haven't had a chance to lay my hands on one yet but seems really cool.
I really like the idea of Linux but I'm not quite ready to give it a shot yet. I feel slightly disgusted whenever I try a Mac.. although Linux would very likely be an easier transition.

etra0:
@Godnoken 
> Thought so. I'm not a big fan of other OSs but Linux isn't out of the question for me in the future as a user..

To be fair, I tried hudhook running on Linux and it works quite well! it can also run in your favorite handheld, which, coincidentally, runs on linux :P

image

Godnoken:
> To be fair, I tried hudhook running on Linux (under wine) and it works quite well! it can also run in your favorite handheld, which, coincidentally, runs on linux :P
> 
> Image related to Hudhook running on Linux

That's awesome! I'll have to get my hands on one of those when money allows 😄 
I wonder if my app would work there.. probably too hard to use on a handheld though.

Jakobzs:
Thanks for the addition here Godnoken, and as per TODO, yes it is definitely warranted 😄. The `gl_loader` repository is still at Rust 2018 and the last commit was 4 years ago so... yeah.

One thing that perhaps could be considered is to follow more the style of `Glad` and have it as a two-step process:
1. [Opening OpenGL, storing the libGL (module) and grabbing `wglGetProcAddress`](https://github.com/maeln/gl_loader/blob/master/src/glad.c#L37)
2. [Getting OpenGL functions](https://github.com/maeln/gl_loader/blob/master/src/glad.c#L111)

Overall I believe this may still be possible to have as one function, but use an initializer like `OnceCell`, perhaps just for storing the module and nothing else, as to not call LoadLibraryW each time. Of course also, this implementation here as discussed above is Windows only... but so is the rest of hudhook (for now) 😛

I have made a few comments in the code as well, just for curiosity mostly.

Godnoken:
> Overall I believe this may still be possible to have as one function, but use an initializer like `OnceCell`, perhaps just for storing the module and nothing else, as to not call LoadLibraryW each time. Of course also, this implementation here as discussed above is Windows only... but so is the rest of hudhook (for now) 😛

I fixed the repeated loading of Opengl32.dll as suggested with OnceCell.
Not sure exactly what you meant with `wglGetProcAddress` as I believe it is necessary to try that first & then go over to `GetProcAddress`. I can't remember why though... Possibly had something to do with 32-bit vs 64-bit?

> I have made a few comments in the code as well, just for curiosity mostly.

Where? 😄 

---

Cheers for the input!

Jakobzs:
> 

Looks like I forgot to submit the comments...

I'll admit I'm not too knowledgeable on this haha. But it does look like it stores the address pointer to `wglGetProcAddress` [here](https://github.com/maeln/gl_loader/blob/master/src/glad.c#L15), although I assume the import you're doing is practically the same thing.

I will just quickly remake some of the comments I had, but have removed those about the "Glad" approach as I think it's very solid now. I just have one comment about the casting though, will remake it.

src/renderers/imgui_opengl3.rs
```diff
@@ -1,3 +1,34 @@
...
+    opengl_wide_string.push(0);
+
+    LoadLibraryW(PCWSTR(opengl_wide_string.as_ptr())).unwrap()
+}
+
+pub unsafe fn get_proc_address(function_string: &str) -> *const c_void {
+    let module = OPENGL3_LIB.get_or_init(|| Mutex::new(get_opengl3_lib())).lock();
+
+    let function_c_string = CString::new(function_string).unwrap();
+    let function_name_ptr = function_c_string.as_bytes_with_nul().as_ptr();
```
Jakobzs, line 24: Is this cast necessary? I believe for most cases where I have used PCSTR, PCSTR(x.as_ptr() as _) or similar was sufficient, given that the function string provided (hopefully) should be null-terminated.

Jakobzs: Mostly just asking if you took this from some other reference or so. Honestly I don't think I've seen it before.

Godnoken: Aha! No, it wasn't required at all. I'm a fresh Rust newbie (new to types as well..) so there's a lot of trial and error haha.
I believe I just looked at other C/C++ implementations + the windows-rs docs when I did this. I really had no idea what I was doing at the time - although I did get it done, albeit not so optimized 😆

Godnoken:
> I'll admit I'm not too knowledgeable on this haha. But it does look like it stores the address pointer to `wglGetProcAddress` [here](https://github.com/maeln/gl_loader/blob/master/src/glad.c#L15), although I assume the import you're doing is practically the same thing.

Oh, interesting. I'll do some quick research and see if it is necessary for an optimized version. Thank you!

src/renderers/imgui_opengl3.rs
Jakobzs: Ah I see, it is fine though 😄

Jakobzs:
> > I'll admit I'm not too knowledgeable on this haha. But it does look like it stores the address pointer to `wglGetProcAddress` [here](https://github.com/maeln/gl_loader/blob/master/src/glad.c#L15), although I assume the import you're doing is practically the same thing.
> 
> Oh, interesting. I'll do some quick research and see if it is necessary for an optimized version. Thank you!

Honestly we're down to micro-optimizations now though. In its current state, it LGTM 👍

Jakobzs (approved):

Jakobzs:
Any reason for LoadLibraryW over LoadLibraryA btw? I am not sure if some systems are forced to run with LoadLibraryW or so, but `opengl32.dll` is ANSI compliant per se, so yea.

Godnoken:
> Any reason for `LoadLibraryW` over `LoadLibraryA` btw? I am not sure if some systems are forced to run with LoadLibraryW or so, but `opengl32.dll` is ANSI compliant per se, so yea.

Nope, I had not fully understood the difference back then so that's why. I'll swap it now.

Godnoken:
> Honestly we're down to micro-optimizations now though. In its current state, it LGTM 👍

Yeah.. don't think there's any optimization to do there after looking at it.

By the way - `wglGetProcAddress` is supposed to work fine on Windows PCs on its own (haven't tried now) and `GetProcAddress` should work universally - but doesn't on my end. Hence why it was written that way. I guess we're halfway to Mac & Linux support on this one 😄

etra0:
> Any reason for `LoadLibraryW` over `LoadLibraryA` btw? I am not sure if some systems are forced to run with LoadLibraryW or so, but `opengl32.dll` is ANSI compliant per se, so yea.

I think we should always prefer LoadLibraryW over the ANSI compliant if we're not using relative paths. As soon as an user has a non english character, it would crash.

Sages0ft:
> > Honestly we're down to micro-optimizations now though. In its current state, it LGTM 👍
> 
> Yeah.. don't think there's any optimization to do there after looking at it.
> 
> By the way - `wglGetProcAddress` is supposed to work fine on Windows PCs on its own (haven't tried now) and `GetProcAddress` should work universally - but doesn't on my end. Hence why it was written that way. I guess we're halfway to Mac & Linux support on this one 😄

Yeah I think you’re right about that.

Godnoken:
> I think we should always prefer LoadLibraryW over the ANSI compliant if we're not using relative paths. As soon as an user has a non english character, it would crash.

Wouldn't be an issue here though, would it? As "opengl32.dll" is hardcoded - I assume the DLL isn't named anything else in other areas.

Jakobzs:
> > Honestly we're down to micro-optimizations now though. In its current state, it LGTM 👍
> 
> Yeah.. don't think there's any optimization to do there after looking at it.
> 
> By the way - `wglGetProcAddress` is supposed to work fine on Windows PCs on its own (haven't tried now) and `GetProcAddress` should work universally - but doesn't on my end. Hence why it was written that way. I guess we're halfway to Mac & Linux support on this one 😄

Definitely looks like it’s possible to move it towards Linux and Mac in the future yep 😄

Jakobzs:
> > I think we should always prefer LoadLibraryW over the ANSI compliant if we're not using relative paths. As soon as an user has a non english character, it would crash.
> 
> Wouldn't be an issue here though, would it? As "opengl32.dll" is hardcoded - I assume the DLL isn't named anything else in other areas.

Fairly sure about this too. I don’t believe it will ever refer to a relative path.

veeenu (changes requested):

veeenu:
> To be fair, I tried hudhook running on Linux (under wine) and it works quite well! it can also run in your favorite handheld, which, coincidentally, runs on linux :P
> Image related to Hudhook running on Linux

I love it!

> Now, natively is another issue, but I don't see value in that since at that point you're better integrating imgui-rs itself, considering most projects on linux are open source.
> 
> The one thing I couldn't manage to do is to compile it inside the Linux box (using the target `x86_64-pc-windows-gnu`). From a very brief exploration it seemed to be related to `OnceCell`, but I didn't dig up too much, too many layers of abstractions.
> 
> Anyway, sorry for the offtopic, just thought it would be worth posting.

No problem! `OnceCell` is still experimental in Nightly so it's unsurprising that it's a bit wobbly. If we ever find the need to make cross compilation work, we can consider using the `once_cell` crate which iirc the `std` implementation is based off of, but still works under stable.

src/renderers/imgui_opengl3.rs
```diff
@@ -1,3 +1,30 @@
...
+    let opengl3_cstring = CString::new("opengl32.dll").unwrap();
+
+    LoadLibraryA(PCSTR(opengl3_cstring.as_ptr() as _)).unwrap()
+}
+
+/// # Safety
+///
+/// Help me out lol
+pub unsafe fn get_proc_address(function_string: CString) -> *const c_void {
+    let module = OPENGL3_LIB.get_or_init(|| Mutex::new(get_opengl3_lib())).lock();
```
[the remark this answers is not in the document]
Godnoken: Probably not, works fine without it!

src/renderers/imgui_opengl3.rs
```diff
@@ -1,3 +1,29 @@
...
-pub use {gl_loader, imgui_opengl_renderer};
+use std::ffi::{c_void, CString};
+
+use once_cell::sync::OnceCell;
+use windows::core::PCSTR;
+use windows::Win32::Foundation::{FARPROC, HINSTANCE};
+use windows::Win32::Graphics::OpenGL::wglGetProcAddress;
+use windows::Win32::System::LibraryLoader::{GetProcAddress, LoadLibraryA};
+
+static mut OPENGL3_LIB: OnceCell<HINSTANCE> = OnceCell::new();
```
veeenu, line 9: Last thing and then we're good to go, `OnceCell::get_or_init` should not require mutability so I think this should be preferable:

```suggestion
static OPENGL3_LIB: OnceCell<HINSTANCE> = OnceCell::new();
```

veeenu (changes requested):

src/renderers/imgui_opengl3.rs
```diff
@@ -1,3 +1,29 @@
...
+
+unsafe fn get_opengl3_lib() -> HINSTANCE {
+    let opengl3_cstring = CString::new("opengl32.dll").unwrap();
+
+    LoadLibraryA(PCSTR(opengl3_cstring.as_ptr() as _)).unwrap()
+}
+
+/// # Safety
+///
+/// Help me out lol
```
Jakobzs, line 19: Probably should change to something relevant here before merging too 😛 idk what though

Jakobzs (changes requested):

veeenu (approved):
LGTM!

PR #81, opened 2023-03-12
Added tracing

Jakobzs:
This PR adds `tracing` as a logger over `log` and `simplelog`, along with a util function for showing color in the console.

Logging is kind of a personal preference in terms of which library to use, but some reasons for choosing `tracing` is that it's hugely more popular, featureful in comparison to the other solutions, and is also managed by the Tokio team. It also helps that the current logging implementation barely requires any changes to adapt to `tracing` 😄


Godnoken:
Out of curiosity, what is the purpose of this change?

tests/dx11.rs
```diff
@@ -54,9 +53,6 @@ fn test_imgui_dx11() {
         }
     }
 
-    TermLogger::init(LevelFilter::Trace, Config::default(), TerminalMode::Mixed, ColorChoice::Auto)
-        .unwrap();
-
```
veeenu, line 59: Can we add a tracing subscriber here? Seeing logs is pretty important for tests (same for other tests).

hudbook/hello-hud/Cargo.toml
```diff
@@ -14,5 +14,5 @@ crate-type = ["cdylib", "rlib"]
 name = "hello_hud"
 
 [dependencies]
-hudhook = { version = "0.4.0", path = "../..", features = ["simplelog"] }
+hudhook = { version = "0.4.0", path = "../..", features = [] }
```
veeenu, line 17:
```suggestion
hudhook = { version = "0.4.0", path = "../.." }
```

examples/dx11_hook.rs
```diff
@@ -9,8 +9,6 @@ impl Dx11HookExample {
     fn new() -> Self {
         println!("Initializing");
         hudhook::utils::alloc_console();
-        #[cfg(feature = "simplelog")]
-        hudhook::utils::simplelog();
```
veeenu, line 13: Same as for the tests below -- can we add a tracing subscriber here?

Cargo.toml
```diff
@@ -99,17 +98,16 @@ windows = { version = "0.39.0", features = [
   "Win32_UI_Input_KeyboardAndMouse",
   "Win32_UI_WindowsAndMessaging",
 ] }
-simplelog = { version = "0.12.0", optional = true }
+tracing = "0.1"
 
 [dev-dependencies]
-simplelog = "0.12.0"
+# tracing-subscriber = "0.3"
```
veeenu, line 104: I'd remove the `dev-dependencies` table entirely if we don't use any of its values; but we may need this anyway (see other comments)

veeenu (changes requested):
Thank you for this! From a cursory check, it looks good in general, but I'd add logging in test/example modules. Haven't had a chance to test it but will do so soon.

Jakobzs:
Tracing subscribers should be added now.

veeenu (changes requested):
I'm a bit scared because it took me quite a while to adapt one practice tool's code to work with tracing, so this will be quite the breaking change. But it seems like it also improved code quality somewhat and I prefer `tracing`'s approach. Once the two modifications I suggested are ready, we can merge.

src/lib.rs
```diff
@@ -232,12 +239,13 @@ pub mod lifecycle {
     }
 }
```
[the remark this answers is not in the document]
Sages0ft: Done

veeenu (approved):

Godnoken:
> It is mainly to integrate a hugely more popular logging framework into hudhook.

Sorry for not getting back to this. Been crunching on my own stuff for a bit.

Interesting. Without having had a look at it - I assume I will no longer be able to see logs from hudhook from my own application if I stick to log and simplelog? I just implemented it a while ago and I'm quite content with it, but if not I'll have to change.

veeenu:
> Interesting. Without having had a look at it - I assume I will no longer be able to see logs from hudhook from my own application if I stick to log and simplelog? I just implemented it a while ago and I'm quite content with it, but if not I'll have to change.

Yep, that's correct. It's quite a big breaking change, but luckily the user facing API is almost 1:1 with tracing. Check [this PR](https://github.com/veeenu/eldenring-practice-tool/pull/46/files) out -- it looks big because it has other unrelated lints, but the change should be pretty small.

Godnoken:
Fair enough, I'll take a look. Cheers!

PR #82, opened 2023-04-01
Added new examples. Modified an API.

vSylva:
Added two dll proxy examples.
Modified an API.

### DIFF
--- a/examples/dinput8_hook/Cargo.toml
+++ b/examples/dinput8_hook/Cargo.toml
@@ -1,0 +1,16 @@
+[package]
+name = "dinput8_hook"
+version = "0.1.0"
+edition = "2021"
+authors = ["AuroraSilvaX", "Andrea Venuta <venutawebdesign@gmail.com>"]
+
+[lib]
+name = "dinput8"
+crate-type = ["cdylib"]
+
+[dependencies]
+windows = { version = "0.39.0", features = [
+    "Win32_Devices_HumanInterfaceDevice",
+    "Win32_Foundation",
+    "Win32_System_LibraryLoader",
+] }
--- a/examples/dinput8_hook/Cargo.toml
+++ b/examples/dinput8_hook/Cargo.toml
@@ -2,7 +2,7 @@
 name = "dinput8_hook"
 version = "0.1.0"
 edition = "2021"
-authors = ["AuroraSilvaX", "Andrea Venuta <venutawebdesign@gmail.com>"]
+authors = ["vSylva"]
 
 [lib]
 name = "dinput8"
--- a/examples/dinput8_hook/src/lib.rs
+++ b/examples/dinput8_hook/src/lib.rs
@@ -1,0 +1,49 @@
+#![feature(once_cell)]
+#![allow(non_snake_case, unused_variables, unreachable_patterns)]
+
+use std::ffi::c_void;
+use std::sync::LazyLock;
+
+use windows::core::{GUID, HRESULT, PCSTR};
+use windows::s;
+use windows::Win32::Foundation::{BOOL, HINSTANCE};
+use windows::Win32::System::LibraryLoader::{GetProcAddress, LoadLibraryA};
+
+static ENTRY_POINT: LazyLock<DirectInput8Create> = LazyLock::new(|| unsafe {
+    let handle = LoadLibraryA(s!("C:\\Windows\\System32\\dinput8.dll\0")).unwrap();
+    std::mem::transmute(GetProcAddress(handle, PCSTR(b"DirectInput8Create\0".as_ptr())).unwrap())
+});
+
+type DirectInput8Create = unsafe extern "stdcall" fn(
+    hinst: HINSTANCE,
+    dwversion: u32,
+    riidltf: *const GUID,
+    ppvout: *mut *mut c_void,
+    punkouter: HINSTANCE,
+) -> HRESULT;
+
+#[no_mangle]
+#[export_name = "DirectInput8Create"]
+unsafe extern "stdcall" fn direct_input8_create(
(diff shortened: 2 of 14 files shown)


examples/dinput8_hook/src/lib.rs
veeenu, line 1: I like what this example is trying to demonstrate, but it is not using `hudhook` at all. I'd prefer if it was a bit more informative, and maybe it would be better suited for the [hudbook](https://veeenu.github.io/hudhook/).

examples/dinput8_hook/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+use std::ffi::c_void;
+use std::sync::LazyLock;
+
+use windows::core::{GUID, HRESULT, PCSTR};
+use windows::s;
+use windows::Win32::Foundation::{BOOL, HINSTANCE};
+use windows::Win32::System::LibraryLoader::{GetProcAddress, LoadLibraryA};
+
+static ENTRY_POINT: LazyLock<DirectInput8Create> = LazyLock::new(|| unsafe {
+    let handle = LoadLibraryA(s!("C:\\Windows\\System32\\dinput8.dll\0")).unwrap();
```
veeenu, line 13: I appreciate hardcoding that path may work on 99% of users' computers, but I think a more general approach would be better here, even if it is more convoluted. This is a good example of what I mean:

https://github.com/veeenu/eldenring-practice-tool/blob/1216fb55c5b8f415ac74dea815dd2717981993b4/lib/no-logo/src/lib.rs#L54-L65

examples/dinput8_hook/Cargo.toml
veeenu, line 1: I'd prefer not to have separate crates in `examples`, and go with the Rust idiom (simply structuring examples in bare subdirectories) instead, so that it can be better managed by Cargo on the top-level. I'm not sure the current solution can work properly by running `cargo run --example dinput8_hook` and it could be confusing to some. 

If we need more dependencies for the examples, we could use the `dev-dependencies` section of the top-level `Cargo.toml`; but even then, I'd prefer keeping those to the absolute minimum.

examples/dinput8_hook/src/lib.rs
```diff
@@ -0,0 +1,49 @@
+#![feature(once_cell)]
+#![allow(non_snake_case, unused_variables, unreachable_patterns)]
```
veeenu, line 2: I'd strongly prefer not to use these. These disabled lints are some of those I deem most important for maintaining high code quality.

examples/dinput8_hook_d3d12/src/lib.rs
veeenu, line 1: Most of the comments I made for the other example file are still valid for this one; but this is a better, more informative, and more pertinent example, so I'd keep this one rather than the other.

examples/dinput8_hook_d3d12/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+            .set_target_level(LevelFilter::Error)
+            .build();
+        TermLogger::init(LevelFilter::Error, config, TerminalMode::Mixed, ColorChoice::Auto).ok();
+
+        Dx12HookExample
+    }
+}
+
+impl ImguiRenderLoop for Dx12HookExample {
+    fn should_block_messages(&self, _io: &mut imgui::Io) -> bool {
```
veeenu, line 37: When a parameter name is actually used, I'd prefer the leading `_` to be removed.
Also, the `Io` argument should _not_ be `&mut` as this method is meant to be a predicate -- see below for a more complete explanation.

examples/dinput8_hook_d3d12/src/lib.rs
```diff
@@ -0,0 +1,114 @@
...
+            _io.mouse_draw_cursor = true;
+            return true;
+        } else {
+            _io.mouse_draw_cursor = false;
+            return false;
+        }
+    }
+
+    fn render(&mut self, ui: &mut imgui::Ui, _: &ImguiRenderLoopFlags) {
+        ui.window("Hello world").size([300.0, 110.0], Condition::FirstUseEver).build(|| {
```
veeenu, line 55: Using `ui.io_mut()` paired with `should_block_messages` here would be more appropriate for assessing whether to draw the cursor or not, and making the relevant modifications.

rust-toolchain.toml
```diff
@@ -1,2 +1,4 @@
 [toolchain]
-channel = "nightly"
+channel = "nightly-2023-03-03"
+
+# Nightly: 1.70.0 (1 June, 2023, 61 days left) does not support #![feature(once_cell)]
```
veeenu, line 4: Good point here -- we're in the process of doing away with nightly altogether, since the newly stabilized `once_cell` should be enough for us to get rid of that.

src/hooks/common.rs
```diff
@@ -217,7 +217,7 @@ pub trait ImguiRenderLoop {
 
     /// If this function returns true, the WndProc function will not call the
     /// procedure of the parent window.
-    fn should_block_messages(&self, _io: &Io) -> bool {
+    fn should_block_messages(&self, _io: &mut Io) -> bool {
```
veeenu, line 220: This is a breaking API change which I'd strongly prefer _not_ having, for a number of reasons:
- Clients already rely on `io` being immutable; changing this would cause a number of compilation issues on projects depending on `hudhook`, and this is bad unless properly communicated and planned for.
- `should_block_messages` sounds like a question asked of the current state; it is not immediately clear to the user why it would mutate state anywhere.
- There are no strongly compelling reasons for such a big behavior change; the client code (examples in this case) should rather adapt.

vSylva:
Thank you for your reply which gave me some tips

PR #90, opened 2023-04-24
Fix DX9 early inject crash

Godnoken:
This PR fixes an early injection crash that I have (knowingly) only experienced in Fallout 3, but it likely happens in others too.

I honestly am not sure why this works, but changing the d3d9 devicetype from `D3DDEVTYPE_HAL` to `D3DDEVTYPE_NULLREF` makes it stop.

I tested it on 10+ DX9 games without experiencing side-effects, so it should be okay to do this.

Below is the crash I get with `D3DDEVTYPE_HAL`;
image

### DIFF
--- a/src/hooks/dx9.rs
+++ b/src/hooks/dx9.rs
@@ -10,7 +10,7 @@ use windows::Win32::Foundation::{
 };
 use windows::Win32::Graphics::Direct3D9::{
     Direct3DCreate9, IDirect3DDevice9, D3DADAPTER_DEFAULT, D3DBACKBUFFER_TYPE_MONO,
-    D3DCREATE_SOFTWARE_VERTEXPROCESSING, D3DDEVTYPE_HAL, D3DDISPLAYMODE, D3DFORMAT,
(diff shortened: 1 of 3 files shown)


veeenu (changes requested):
Thank you!

I wonder whether a similar approach could work for other DX renderers as well. I fiddled with device types but without a whole lot of success. Might give it a crack again.

Barring the cargo fmt, LGTM!

Godnoken:
I haven't noticed any crashes with DX11 (I don't have any DX12 games..).
Any specific games you have had problems with?

veeenu:
> I haven't noticed any crashes with DX11 (I don't have any DX12 games..). Any specific games you have had problems with?

Both DS3 and ER 😅 I've had this recurring problem of occasional crashes when the tools are injected too early for forever. Been toying with different kinds of devices but never found a good solution yet.

veeenu (approved):

Godnoken:
Ahh! I have neither of those games unfortunately.. Altough I've probably had the same crash but never noticed before.

PR #93, opened 2023-04-26
Add tracing feature 'log'

Godnoken:
We just need to add the feature "log" to tracing and it will automatically use log if no tracing subscriber is active. More info [here](https://docs.rs/tracing/0.1.29/tracing/index.html#log-compatibilityl).

Thoughts? Works flawlessly :)


Jakobzs:
I think this could be a good idea, looking at what is mentioned [here](https://docs.rs/tracing/latest/tracing/index.html#crate-feature-flags) as well. I did not know that feature flag exists, good job.

veeenu (dismissed):
I LOVE this. 💯 

LGTM, can merge in as soon as `cargo fmt` is run.

Issue #36, opened 2022-10-27
Add a way to display the mouse and lock keyboard input.


etra0:
I'm still not totally sure about this but it would be worth the discussion.

Currently, although this captures mouse input, it doesn't have any mechanism to 'force' the cursor to be displayed.
I'd like to be something like Reshade, when the Menu is open, show and capture mouse input (and blocks mouse input
from the game).

Right now I'm doing some *very* nasty hack for the tool I've been building for The Witcher and it kinda works,
it shows the mouse even if the game doesn't want to:

https://github.com/etra0/hudhook/compare/master...temp_patches

This could work as a proof of concept I guess if we test it in more games.

veeenu:
I used to have something similar way back when and I think it would be a good idea, because some games will definitely need it.

I don't think your solution is particularly hacky, it's definitely the "intended" way of showing a cursor. The main thing I'd change is that the decision to show the cursor should be delegated to the clients via some API. I'm not sure what that could look like yet.

etra0:
The solution is actually using [`ImguiRenderLoop::should_block_messages`](https://github.com/veeenu/hudhook/blob/d8f18f4dd3c0535d459f104527ff705f849b6b54/src/hooks/common.rs#L181) so it's up to the client to re-implement that trait in order to show the cursor. Maybe could be renamed

etra0:
I'd also like a way for the mouse input to be totally blocked, that would require a bit of investigation, because sometimes when you click you also interact with the game and it's a bit messy. Right now my hack only displays the cursor

veeenu:
I've kind of given up on that. Been trying to block input for DS3 and ER for _years_ now, to absolutely no avail. Still haven't understood how to block their input flow, I've no idea where they even get the data from.

etra0:
Good thing is reshade is also open source :D I'll try some stuff when I have the time and see if a PR is possible

Godnoken:
I'm also very interested in finding a way to block game input completely, in all games. I'm going to investigate it over the next month if time allows.

> Good thing is reshade is also open source :D I'll try some stuff when I have the time and see if a PR is possible

Are you saying that Reshade implements this behaviour?

etra0:
iirc yes it does, but I may be wrong. It's been a while since I had time to look this up.

Godnoken:
Interesting. I'm going to give it a download and see if it works for a wide variety of games.

Godnoken:
It seems to work almost flawlessly, amazing.

[...]

PR #94, opened 2023-04-26
Revamp input handling

Godnoken:
As shown in https://github.com/veeenu/hudhook/issues/36#issuecomment-1519061031, I have found a way to block all input in most games by studying OSS Reshade and a bunch of forums.

Current issues;

- When `should_block_messages`  is set to true for longer periods of time (?), game input stops being received altogether in some games. Seems to be of a greater issue in games that use Raw input. ALT + TAB *may* fix it.
- Some games still catch mouse hover actions - no clue why as of right now.
- The in-game cursor will still follow the IMGUI cursor in a lot of newer games. I'm not confident that this is something we can solve.
- We are currently not blocking games that use DirectInput, but this should be easily solved by hooking into it.
- All hooks, especially all new hooks that aren't `SetWindowsHookEx` need to be heavily tested & some may be completely unnecessary (most likely Post/Get/Peek-Message).
- Make my code idiomatic! 🥇 

Any questions, help or thoughts on this is very much appreciated. I'm just a noob brute-forcing my way to success.


Godnoken:
> * When `should_block_messages`  is set to true for longer periods of time (?), game input stops being received altogether in some games. Seems to be of a greater issue in games that use Raw input. ALT + TAB _may_ fix it.

I have with great confidence tracked this down to being relevant only to games that use Raw input now. Should be able to fix this soon.

veeenu:
Thank you for doing this work! I feel like this is going in the right direction, though I would address a few architectural shortcomings:

- having many instances of `OnceCell` and `std::mem::transmute` -- there are probably better patterns to achieve the same results, like putting everything into a single struct and working better with types; `transmute` should in general be the absolute last resort and heavily scrutinized and contained when used.
- passing pointers into functions: I'd rather have one single spot for each pointer where it gets validated and turned into a reference with `.as_ref()`, this improves _dramatically_ code readability, does not require `unsafe` for functions, naturally guides you into better structuring and subdividing your code.
- I'd also like to reduce hooking to the absolute minimum -- we should investigate whether we can just find other `HWND`s of the game and set our wnd proc there instead of the current solution of hooking into the existing wnd proc. Hooking into an existing function _works_, of course, but it's more undefined behavior we have to support that I'd avoid like the plague unless we're absolutely, positively, 100% sure there is no better way possible at all. The maintenance burden with the current hooks, which are absolutely necessary for the library to function, is already pretty bad due to voluntarily stepping into undefined behavior; and it is bound to grow exponentially with every single hook added. So I'd like to do the absolute best possible to avoid ever adding a new hook again. Honestly, I'd happily choose to go with fewer features over adding more hooks.

Overall, the input handling code looks good and way more nuanced than what we have currently, so I'd gladly integrate this once it gets in a good state.

Thank you again!

Godnoken:
> * having many instances of `OnceCell` and `std::mem::transmute` -- there are probably better patterns to achieve the same results, like putting everything into a single struct and working better with types; `transmute` should in general be the absolute last resort and heavily scrutinized and contained when used.

I agree fully and will sort this out shortly.

> * passing pointers into functions: I'd rather have one single spot for each pointer where it gets validated and turned into a reference with `.as_ref()`, this improves _dramatically_ code readability, does not require `unsafe` for functions, naturally guides you into better structuring and subdividing your code.

Sure thing, I'll study some of the hudhook code already in place and adjust.

> * I'd also like to reduce hooking to the absolute minimum -- we should investigate whether we can just find other `HWND`s of the game and set our wnd proc there instead of the current solution of hooking into the existing wnd proc. 

Hmm. I'm not sure what you mean here. I believe that the only way to successfully block all game input is to hook into the message handlers being used one way or another. 

> * The maintenance burden with the current hooks, which are absolutely necessary for the library to function, is already pretty bad due to voluntarily stepping into undefined behavior; and it is bound to grow exponentially with every single hook added. So I'd like to do the absolute best possible to avoid ever adding a new hook again. Honestly, I'd happily choose to go with fewer features rather than adding more hooks if those were the only two options possible.

Oof.
Unfortunately, there seems to be a huge issue with mouse raw input suddenly dying after continuously blocking. I'm not 100% sure that it is because we're using `SetWindowsHookEx` yet. However, I think I have figured out how Reshade manages to hook into all the functions needed to successfully block. We could go down that approach, but it would require us to use all the hooks in this PR (the hooks aren't complete, nor are most of them even used currently) and then some.

I understand that you want to be cautious with this but I'm starting to doubt that this mouse raw input issue is going to be fixed without hooking into the specific functions. I've been at it for two days now without any progress. At some point the game just decides to stop receiving any mouse input.

I wouldn't have any issue if you don't want to take that direction, besides, games that don't use raw input for the mouse would work just fine with just this one hook and currently Hudhook doesn't even have raw input handling.

I'll make another post soon about what I've tried and where I believe it crashes (?).

> Overall, the input handling code looks good and way more nuanced than what we have currently, so I'd gladly integrate this once it gets in a good state.
> 
> Thank you again!

Thank you!

Godnoken:
So.

The input freeze seems to be a fairly documented issue, granted that it is the same one that we are suffering currently.

Some threads about the problem(s);
[WH_MOUSE_LL](https://social.msdn.microsoft.com/Forums/windowsdesktop/en-US/f6032ca1-31b8-4ad5-be39-f78dd29952da/hooking-problem-in-windows-7?forum=windowscompatibility)
[WH_KEYBOARD_LL](https://stackoverflow.com/questions/1639331/using-global-keyboard-hook-wh-keyboard-ll-in-wpf-c-sharp)
[WH_MOUSE](https://stackoverflow.com/questions/20734671/global-mouse-hook-causes-application-crash-c-windows7)
[WH_MOUSE_LL](https://stackoverflow.com/questions/12401513/global-mouse-hook-suddenly-not-getting-called)

The ones with _LL at the end are desktop global hooks, however, they seem to be suffering from the same issues nonetheless.

- Solutions seem to have ranged from extending the timeout on a registry key called `LowLevelHooksTimeout`;


[MSDN](https://social.msdn.microsoft.com/Forums/windowsdesktop/en-US/f6032ca1-31b8-4ad5-be39-f78dd29952da/hooking-problem-in-windows-7?forum=windowscompatibility#:~:text=We%20have%20just%20found%20this%20in%20our%20application.%20We%20added%20this%20DWORD%20to%20the%20registry%20and%20the%20problem%20went%20away.%0A%0AHKEY_CURRENT_USER%5CControl%20Panel%5CDesktop%5C%0A%0ALowLevelHooksTimeout%3D10000)


- Create a new thread, hook from there and then create a message loop to keep pumping out messages and keep thread alive;

```DWORD WINAPI mouseLLHookThreadProc(LPVOID lParam)
{
    MSG msg;

    _hMouseLLHook = SetWindowsHookEx( WH_MOUSE_LL, .....);

    while(GetMessage(&msg, NULL, 0, 0) != FALSE)
    {
        TranslateMessage(&msg);
        DispatchMessage(&msg);
    }

    return 0;
}
```
More on that from [MSDN](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/ms644986(v=vs.85)#:~:text=This%20hook%20is%20called%20in%20the%20context%20of%20the%20thread%20that%20installed%20it.%20The%20call%20is%20made%20by%20sending%20a%20message%20to%20the%20thread%20that%20installed%20the%20hook.%20Therefore%2C%20the%20thread%20that%20installed%20the%20hook%20must%20have%20a%20message%20loop.) (only specifies _LL here) and [StackOverflow](https://stackoverflow.com/questions/20734671/global-mouse-hook-causes-application-crash-c-windows7#:~:text=user2802841%20was%20right%2C%20the%20problem%20was%20the%20missing%20message%20loop.%20When%20using%20a%20mouse%20hook%20(WH_MOUSE%20or%20WH_MOUSE_LL)%20in%20a%20console%20application%20getMessage%20or%20peekMessage%20have%20to%20be%20called%20somewhere%20in%20the%20application.)


- Rehooking every so often to not get timed out;

[StackOverflow](https://stackoverflow.com/questions/1639331/using-global-keyboard-hook-wh-keyboard-ll-in-wpf-c-sharp#:~:text=I%20have%20used%20the%20Dylan%27s%20method%20to%20hook%20global%20keyword%20in%20WPF%20application%20and%20refresh%20hook%20after%20each%20key%20press%20to%20prevent%20events%20stop%20firing%20after%20few%20clicks%20.%20IDK%2C%20if%20it%20is%20good%20or%20bad%20practice%20but%20gets%20the%20job%20done.)


- Seems irrelevant to Rust but here's someone talking about garbage collection;

[StackOverflow](https://stackoverflow.com/questions/1639331/using-global-keyboard-hook-wh-keyboard-ll-in-wpf-c-sharp#:~:text=You%27re%20creating%20your%20callback%20delegate%20inline%20in%20the%20SetHook%20method%20call.%20That%20delegate%20will%20eventually%20get%20garbage%20collected%2C%20since%20you%27re%20not%20keeping%20a%20reference%20to%20it%20anywhere.%20And%20once%20the%20delegate%20is%20garbage%20collected%2C%20you%20will%20not%20get%20any%20more%20callbacks.)

---

I tried all solutions above and there's likely more that I have forgotten, but it seems that using `SetWindowsHookEx` is less than ideal in some cases..

I also tried;

- Only hooking into the thread that uses raw input
- Use [CallNextHookEx](https://learn.microsoft.com/en-us/previous-versions/windows/desktop/legacy/ms644981(v=vs.85)#:~:text=If%20code%20is%20greater,value%20should%20be%20zero.) in various ways
- Hooking from `DLL_PROCESS_ATTACH` with and without a new thread


The issue has so far only popped up for me if we change raw mouse movement input to `WM_NULL` and then proceed to block it for 7-10 seconds while moving the cursor fast continuously (moving slow seems fine...)
As far as I know, `WH_GETMESSAGE`, `WH_MOUSE` etc get called before the message is sent to `GetMessage` and `PeekMessage`. Hence I believe that when we modify the raw mouse movement to `WM_NULL` we are somehow overloading or "destroying" the game's message loop. However, that theory falls flat since hooking into `GetMessage/PeekMessage` essentially would be no different, apart from the issues listed above.


I'm getting really tired of troubleshooting this. If no one knows something that could help, I suggest that we drop raw mouse input blocking until a solution is found - OR I delve into hooking the functions specifically and hope that the problem lies with the `SetWindowsHookEx` hooks.

Godnoken:
> * having many instances of `OnceCell` -- there are probably better patterns to achieve the same results, like putting everything into a single struct

Fixed this to the best of my ability.

I assume Atomics inside of the Input struct isn't necessary at all, but I just stuck with them because the syntax is so nice. Not sure if it has any performance impact?

Godnoken:
I tried hooking into `Peek/Get/Post-Message` properly without `SetWindowsHookEx` & `WH_GETMESSAGE` and there have so far been zero issues. Blocking input seems to work flawlessly and there is no longer any game input freezing happening.

`SetWindowsHookEx` just doesn't seem to quite cut it. It's 100% plausible that I've done something wrong when using it, but until we figure out what it is.. either we ignore raw input for hudhook or we go down the route of hooking to these functions individually. Either is fine by me, it won't add much work, if even any. If we go with the first option I'll just have a different local fork for the latter since it is absolutely essential for my app to work the way I want it to.

veeenu:
Man, that's a treasure trove of information 😄

Okay so I read a bit more about `SetWindowsHookEx` and it seems it's doing _global_ hooking? Plus it also looks like the timeout behavior is deliberate, and designed to protect from hooks running too long, which makes sense. I'm not liking `SetWindowsHookEx` much either, having considered this.

Btw, I've gone back and checked the code a bit. I'm still very worried about the hooks -- mostly because they run on very fundamental bits of the windows event loop which are going to run a bazillion times per second... Ngl, the thought of adding overhead to every single one of those calls (even if it's just a `call` + `ret`) makes me uncomfortable. The render loop has a cozy 16.6ms to play with at every turn, whereas those might be fired super fast and delay in there is most likely going to mess with the games in unexpected manners -- see the above reason wrt `SetWindowsHookEx`. 

Due to its usage in the practice tools, one of the core needs that I had while designing hudhook was to mess with the game _as little as possible_ in order to not influence the underlying execution in any manner that could be unpredictable. This is already super hard to juggle as is. Integrating this PR in its current state would mean forgoing that design principle altogether.

Don't get me wrong, I think this is super useful and I'd be glad to integrate it in some capacity, but I think it would be worth it to change its design so that it is opt-in -- i.e. an extra set of hooks that you can add at the beginning if you so wish, and that leaves everything else the same.

It would be even better if we found a totally different way that didn't require hooking, or that drastically limited hooking; but I can work with the opt-in thing until we understand the problem space better and find something better.

This might tie in with the idea to expose some internals of the library in order to build external plugins. I want to allocate some cycles to designing that, but lately it's been quite hard for me to do any progress at all due to real life commitments (and the need to unwind a bit).

(as always, thank you so much for all this work!)

>I assume Atomics inside of the Input struct isn't necessary at all, but I just stuck with them because the syntax is so nice. Not sure if it has any performance impact?

IIRC loads and stores should compile down to a single operation anyway on x86/x64, and their only usage is instruction reordering. Should be completely fine 😅

veeenu:
>I tried hooking into Peek/Get/Post-Message properly without SetWindowsHookEx & WH_GETMESSAGE and there have so far been zero issues. Blocking input seems to work flawlessly and there is no longer any game input freezing happening.

Is there a chance that hooking _only those_ could be enough to block the input?

veeenu:
Went and checked Reshade out of curiosity... Found this: https://github.com/crosire/reshade/blob/5ad47a818e919e25db08af94f31abe3744f0bbcc/res/exports.def#L452

So what I did overlook about Reshade is that it's actually a proxy DLL: hooking in that case might be a bit different because it happens at startup time. The address that the host application _sees in the first place_ is the hooked function, so there's no altering of the control flow instructions iiuc. Just find the original function and call it. That's a bit different than what we (are forced to) have. I would still be a bit scared of going that way, honestly, but I can see how that might do the trick (then again maybe the design goals of Reshade are different).

Godnoken:
> * having many instances of `OnceCell` and `std::mem::transmute` -- there are probably better patterns to achieve the same results, like putting everything into a single struct and working better with types; `transmute` should in general be the absolute last resort and heavily scrutinized and contained when used.

I tried alleviating this in the latest commit. However, I don't know if there is any safe way to do the same as transmute for function pointers at the moment?

veeenu:
> I tried alleviating this in the latest commit. However, I don't know if there is any safe way to do the same as transmute for function pointers at the moment?

This looks already significantly better, thank you!

There are things from this PR that I'd like to integrate in isolation, because I think they can stand on their own. In particular, the `handle_window_message` function.

What I'd like to do is start a fresh branch with only that, make a few changes to it (e.g. pass a `&mut Io` to it, retain the wnd proc mechanism via `SetWindowLong`, have the window procs lock the renderers and invoke `handle_window_message` over the `Io` stored in the imgui context) and merge that in once we're satisfied.

Then, we can rebase this PR on main, which will reduce the size of the diff significantly and could pave the way to new ideas and approaches.

Let me know what you think! Would you like me to help with that, or would you prefer taking a crack at it yourself?

Godnoken:
> What I'd like to do is start a fresh branch with only that, make a few changes to it (e.g. pass a `&mut Io` to it, retain the wnd proc mechanism via `SetWindowLong`, have the window procs lock the renderers and invoke `handle_window_message` over the `Io` stored in the imgui context) and merge that in once we're satisfied.

Hmm, but wouldn't this approach make it really clunky to integrate the other hooks? Or do you mean that we would make two different functions like I suggested above, but only integrate the wnd_proc one to begin with?

> Let me know what you think! Would you like me to help with that, or would you prefer taking a crack at it yourself?

I can do it when I understand 😆

veeenu:
> Hmm, but wouldn't this approach make it really clunky to integrate the other hooks? Or do you mean that we would make two different functions like I suggested above, but only integrate the wnd_proc one to begin with?

I mean to integrate things gradually and have as much decoupling as possible so that both approaches can eventually coexist. I wouldn't want to use the multiple hooks approach myself, but I also think the work you have done on them is really valuable; figuring out a way of having them both would furthermore have the advantage of forcing us to rethink a cleaner and more modular API that plays well into the long term plan of extensibility.

This is not trivial, but I think it's the best course of action.

veeenu:
> Understandable, however, I do not see the need to couple the current `handle_window_message` with `wnd_proc`. I believe it would be a lot easier and better for modularity to keep it separate and only pass `umsg` to it. Unless I'm missing something?

Oh, I didn't mean to do that. It's a good idea that the function stands on its own, I'm just going to keep using the current window proc approach (subclassing via `SetWindowLong`) with the new implementation. That function will be wrapped in a wnd proc so we can avoid straying too far from the current implementation all in one go, and test the new input functionality in isolation to make sure it doesn't break clients. Once we're confident of that, we can move on to figure out a solution for the multihooks thing, and simply invoke `handle_window_message` from the hook instead of from the wnd proc wrapper.

veeenu:
I finally had the time to test this branch against Dark Souls III and Elden Ring. I used the branch as-is. Unfortunately, the inputs that I receive are duplicated in both games for some reason, and two cursor are actually shown. The underlying input is indeed blocked, though, which is a great result. 

I wish there was a way of doing this that didn't involve all these hooks -- I have a poor understanding of them and I'm afraid rerouting core Windows events functionality might result in unintended side effects like these.

I just want to thank you again and reiterate that the research and effort you put into this is amazing. Unfortunately we can't merge it currently, but I feel like if we push more and seek to understand the situation better, the day when we will crack will come sooner rather than later!

[out.webm](https://github.com/veeenu/hudhook/assets/2186168/891eb91d-2eae-4397-9c3c-90c89921e5ae)

Godnoken:
> I finally had the time to test this branch against Dark Souls III and Elden Ring. I used the branch as-is. Unfortunately, the inputs that I receive are duplicated in both games for some reason, and two cursor are actually shown. The underlying input is indeed blocked, though, which is a great result.

Interesting. The duplicated input is something I never experienced myself, but neither would I probably notice because it wouldn't make any difference on my app.. 😄 
There are plenty of mentions of it and there are fixes for it in the Reshade codebase, but I never implemented any of it due to not noticing the issues and the fact that it seems fairly complicated.

As for the cursor, you have to put ``SetCursor(HCURSOR(0))`` [here](https://github.com/veeenu/hudhook/pull/94/files#diff-226af8efab9e26fab8d0dab601e863e0e6ed1d7c289868ef57385ba37a36e94dR179-R183) and possibly also hook the ``SetCursor`` function, it depends on the game.

If any game you try likes to lock the cursor somewhere you can also put ``ClipCursor(std::ptr::null())`` in there.

> I just want to thank you again and reiterate that the research and effort you put into this is amazing. Unfortunately we can't merge it currently, but I feel like if we push more and seek to understand the situation better, the day when we will crack will come sooner rather than later!

Thank you very much! Don't worry about it, all the work I put in is also massively benefitting me in many ways. No shame in saying no, quite the opposite - I'm glad to be able to discuss this and get viewpoints from someone with far more experience. You've been very patient with me so far and I appreciate that tremendously.

Godnoken:
Oh, and by the way - I found a way to fix [this](https://github.com/veeenu/hudhook/pull/94#:~:text=may%20fix%20it.-,Some%20games%20still%20catch%20mouse%20hover%20actions%20%2D%20no%20clue%20why%20as%20of%20right%20now.,-The%20in%2Dgame).

It works in all games but one that I have tried so far. It is Star Wars: Empire at War and I currently don't have the faintest idea how the game is registering where the mouse is hovering. It could be DirectInput, an internal input manager or something else. Don't know.

PR #95, opened 2023-05-05
Fix OpenGL resizing

Godnoken:
This PR will fix ImGui not resetting & thus disappearing in some games when a resize, resolution change or ALT + TAB happens.

We have to compare resolution because some older games (like CS 1.6) don't trigger any window resize when it is in fullscreen and the user changes resolution.

We have to compare window rect because ALT + TAB'ing and changing from windowed to/from fullscreen doesn't always trigger resolution change.

---

The only issue I have noticed so far is that after a bunch of resizes it is possible that everything ImGui renders goes completely black. If you resize again it is possible that it disappears.
I assume there is something more needed for the reset on the OpenGL end - maybe `glClearColor`?

image

### DIFF
--- a/src/hooks/opengl3.rs
+++ b/src/hooks/opengl3.rs
@@ -65,6 +65,7 @@ unsafe fn draw(dc: HDC) {
                 wnd_proc,
                 flags: ImguiRenderLoopFlags { focused: false },
                 game_hwnd: hwnd,
+                resolution_and_rect: None,
             };
 
             // Initialize window events on the imgui renderer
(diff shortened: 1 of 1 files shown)


Godnoken:
Fixed everything turning black with `glClearColor`.

*Almost* tricked myself that it wasn't working because I kept rebuilding for x64 when the games I were trying it on are x86... 🤕

src/hooks/opengl3.rs
```diff
@@ -124,6 +128,33 @@ unsafe extern "system" fn imgui_opengl32_wglSwapBuffers_impl(dc: HDC) {
     trampoline_wglswapbuffers(dc)
 }
 
+static mut RESOLUTION_AND_RECT: Option<Mutex<([i32; 2], RECT)>> = None;
```
veeenu, line 131: I'd avoid adding global statics if possible. Global state is hard to track.

I think this instance could be added to the `ImguiRenderer` struct instead, skipping the mutex. The `reset` function could then be changed to lock `IMGUI_RENDERER`, pull this resolution data out of it, determine whether a clean needs to happen, then do a cleanup and finally `IMGUI_RENDERER.take()`. 

This flow is a bit more complicated, but it does catch a potential concurrency error where `IMGUI_RENDERER` is `None` but `cleanup` hasn't happened yet.

veeenu (changes requested):

veeenu (approved):
LGTM! Looks clean and consistent now. We can merge into main after a rebase.

On an unrelated note, I just noticed that `IMGUI_RENDER` is an `Option<T>` instead of `OnceCell` -- I might want to change that to be more consistent across the board and to not accidentally run into concurrency issues (I'm not sure what could be some side effects of using one or the other, but I do know). Can't wait for [this](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) to land into stable and do a pass over all of the other renderers too.

Godnoken:
> On an unrelated note, I just noticed that `IMGUI_RENDER` is an `Option<T>` instead of `OnceCell` -- I might want to change that to be more consistent across the board and to not accidentally run into concurrency issues (I'm not sure what could be some side effects of using one or the other, but I do know). Can't wait for [this](https://doc.rust-lang.org/std/sync/struct.OnceLock.html) to land into stable and do a pass over all of the other renderers too.

Yup, noticed that. There are quite a few inconsistencies across the renderers and hooks, not that it is surprising, but definitely may be worth looking at more closely.

Didn't even notice that OnceCell isn't thread-safe.

veeenu:
> Yup, noticed that. There are quite a few inconsistencies across the renderers and hooks, not that it is surprising, but definitely may be worth looking at more closely.

Yeah, now that I have quite a bit of experience with renderers and hooks under my belt I want to do a full refactor of all of that code eventually, likely once we're able to have more confidence in `windows-rs` again (as per my comments on #98).

> Didn't even notice that OnceCell isn't thread-safe.

It's not a big deal for us since currently we either use that with immutable data structures or wrap `Mutex`es in there, but the fewer data types the better, in my opinion, and switching to `OnceLock` is something I see as desirable eventually as then you don't have to manage that stuff manually.

PR #98, opened 2023-05-15
Bump windows-rs to 0.48.0

Godnoken:
Hope this is alright, not sure if you wanted to do it yourself @veeenu but I need a bump for some DirectX stuff on my end.

Everything works apart from DX11 and I do at the moment not have much of a clue what I did wrong. Still looking into it but I'd appreciate a quick lookover on it!

``Message: panicked at 'called `Result::unwrap()` on an `Err` value: Error { code: HRESULT(0x00000000), message: "The operation completed successfully." }', C:\Users\Godnoken\Desktop\Skrivbord\Programming\hudhook\src\renderers\imgui_dx11\state_backup.rs:95:58``

This is the error I'm getting. Tried fixing that in the last commit (got it from `ctx.RSGetState()` before, and now `ctx.IAGetInputLayout()`).
I assume that these two functions always returned `None` but that it is only now crashing because there wasn't any error handling implemented into them in previous versions of windows-rs.

Either way, I implemented the same error handling for the InputLayout but there is still some obscure crash happening and I don't get any more crash messages..

### DIFF
--- a/src/hooks/dx11.rs
+++ b/src/hooks/dx11.rs
@@ -168,7 +168,7 @@ impl ImguiRenderer {
         trace!("Initializing renderer");
 
         let dev: ID3D11Device = swap_chain.GetDevice().expect("GetDevice");
-        let mut dev_ctx: ID3D11DeviceContext = dev.GetImmediateContext().unwrap();
+        let dev_ctx: ID3D11DeviceContext = dev.GetImmediateContext().unwrap();
 
         let dev_ctx = dev_ctx;
         let mut sd = DXGI_SWAP_CHAIN_DESC::default();
--- a/src/hooks/dx11.rs
+++ b/src/hooks/dx11.rs
@@ -1,6 +1,5 @@
 use std::ffi::c_void;
 use std::mem;
-use std::ptr::null_mut;
 
 use imgui::Context;
 use once_cell::sync::OnceCell;
@@ -169,10 +168,11 @@ impl ImguiRenderer {
         trace!("Initializing renderer");
 
         let dev: ID3D11Device = swap_chain.GetDevice().expect("GetDevice");
-        let mut dev_ctx: Option<ID3D11DeviceContext> = None;
-        dev.GetImmediateContext(&mut dev_ctx);
-        let dev_ctx = dev_ctx.unwrap();
-        let sd = swap_chain.GetDesc().expect("GetDesc");
+        let mut dev_ctx: ID3D11DeviceContext = dev.GetImmediateContext().unwrap();
+
+        let dev_ctx = dev_ctx;
+        let mut sd = DXGI_SWAP_CHAIN_DESC::default();
+        swap_chain.GetDesc(&mut sd).expect("GetDesc");
 
         let engine =
             imgui_dx11::RenderEngine::new_with_ptrs(dev, dev_ctx, swap_chain.clone(), &mut ctx);
@@ -203,7 +203,8 @@ impl ImguiRenderer {
         trace!("Present impl: Rendering");
 
         let swap_chain = self.store_swap_chain(swap_chain);
-        let sd = swap_chain.GetDesc().expect("GetDesc");
+        let mut sd = DXGI_SWAP_CHAIN_DESC::default();
+        swap_chain.GetDesc(&mut sd).expect("GetDesc");
 
         // if GetWindowRect(sd.OutputWindow, &mut rect as _).as_bool() {
         if let Some(rect) = self.engine.get_window_rect() {
@@ -250,7 +251,8 @@ impl ImguiRenderer {
 
     unsafe fn cleanup(&mut self, swap_chain: Option<IDXGISwapChain>) {
         let swap_chain = self.store_swap_chain(swap_chain);
-        let desc = swap_chain.GetDesc().unwrap();
+        let mut desc = DXGI_SWAP_CHAIN_DESC::default();
+        swap_chain.GetDesc(&mut desc).expect("GetDesc");
 
         #[cfg(any(target_arch = "aarch64", target_arch = "x86_64"))]
         SetWindowLongPtrA(desc.OutputWindow, GWLP_WNDPROC, self.wnd_proc as usize as isize);
@@ -312,9 +314,9 @@ fn get_present_addr() -> (DXGISwapChainPresentType, DXGISwapChainResizeBuffersTy
             D3D_DRIVER_TYPE_NULL,
             None,
             D3D11_CREATE_DEVICE_FLAG(0),
-            &[D3D_FEATURE_LEVEL_10_0, D3D_FEATURE_LEVEL_11_0],
+            Some(&[D3D_FEATURE_LEVEL_10_0, D3D_FEATURE_LEVEL_11_0]),
             D3D11_SDK_VERSION,
-            &DXGI_SWAP_CHAIN_DESC {
+            Some(&DXGI_SWAP_CHAIN_DESC {
                 BufferDesc: DXGI_MODE_DESC {
                     Format: DXGI_FORMAT_R8G8B8A8_UNORM,
                     ScanlineOrdering: DXGI_MODE_SCANLINE_ORDER_UNSPECIFIED,
@@ -328,11 +330,11 @@ fn get_present_addr() -> (DXGISwapChainPresentType, DXGISwapChainResizeBuffersTy
                 SwapEffect: DXGI_SWAP_EFFECT_DISCARD,
                 SampleDesc: DXGI_SAMPLE_DESC { Count: 1, ..Default::default() },
                 ..Default::default()
-            },
-            &mut p_swap_chain,
-            &mut p_device,
-            null_mut(),
-            &mut p_context,
+            }),
+            Some(&mut p_swap_chain),
+            Some(&mut p_device),
+            None,
+            Some(&mut p_context),
         )
         .expect("D3D11CreateDeviceAndSwapChain failed");
     }
--- a/src/renderers/imgui_dx11/buffers.rs
+++ b/src/renderers/imgui_dx11/buffers.rs
@@ -1,9 +1,7 @@
-use std::ptr::null_mut;
-
 use imgui::{DrawListIterator, DrawVert};
 use windows::Win32::Graphics::Direct3D11::{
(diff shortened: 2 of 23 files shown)


Godnoken:
`d3d11.dll caused an Access Violation (0xc0000005)
  in module d3d11.dll at 0033:e1688ecb.`

Getting this error from 'Niche' game's dump.

Clearly accessing something incorrectly..

veeenu:
I have consciously decided to stick with `0.39` for `hudhook` because over the last few releases `windows-rs` has gotten _way way worse_ in terms of idiomatics and ergonomics, while no new APIs were added. 

So much so that I had like 3 PRs at the ready over 3 separate repos and decided to close all of them because `0.48` would not only have breaking changes for _all_ my other client code, but also seemingly trivial changes would result in untraceable bugs and crashes.

Not unlike the same issue you mention, which should in principle be easily solvable just by looking at the code side by side with the API docs.

The day `windows-rs` becomes again an actual idiomatic mapping around the Windows API and appropriately leverages the RAII patterns and pointer-like structures, instead of adding counterintuitive undocumented breaking changes, I'll resume with keeping its updates aligned with `hudhook`, but until then, I've decided not to do that because it would block 100% of my work on the dependents of `hudhook`.

I agree with you that it's unfortunate and I would gladly stay up to date for the sake of the clients, but I haven't found a way so far that wouldn't entail giving myself a headache for weeks on end. 

We're already working in a context where bugs show up every now and then that are difficult or impossible to reason about, and that's with very stable and battle tested backends. I have zero confidence in what are the side effects the bump introduces, and I fully expect that if I released a new version of hudhook and the practice tools with `0.48`, I'd be hit with a deluge of bug reports that I have next to zero control over, would cost a lot of time on my end, and would leave people dissatisfied for a long time as I work my way through the reports.

veeenu:
If you need DX12 stuff, using `windows-sys` directly would probably be your best bet as a pointer in there is a pointer in the MS docs; whereas in `windows-rs` currently `Option<T>` might just as well be `T*` in a module`T**` in another, and a straight `T` in yet another, with zero consistency and little relationship to the docs.

src/hooks/dx11.rs
```diff
@@ -169,10 +168,11 @@ impl ImguiRenderer {
...
         let dev: ID3D11Device = swap_chain.GetDevice().expect("GetDevice");
-        let mut dev_ctx: Option<ID3D11DeviceContext> = None;
-        dev.GetImmediateContext(&mut dev_ctx);
-        let dev_ctx = dev_ctx.unwrap();
-        let sd = swap_chain.GetDesc().expect("GetDesc");
+        let dev_ctx: ID3D11DeviceContext = dev.GetImmediateContext().unwrap();
+
+        let dev_ctx = dev_ctx;
+        let mut sd = DXGI_SWAP_CHAIN_DESC::default();
+        swap_chain.GetDesc(&mut sd).expect("GetDesc");
```
veeenu, line 175: Before we had to pass `&mut dev_ctx` as out parameter and get `sd` by return, now we have to do the exact opposite and get `dev_ctx` as return value and pass `&mut sd` as out parameter. I can't find an explanation as to why this is the case.

Of course the absolutely obvious, consistent and ideal thing that we should be able to do here would be:

```
let dev_ctx = dev.GetImmediateContext()?;
let sd = swap_chain.GetDesc()?;
```

I have no idea what this change entails, besides moving the code from "not compiles" to "compiles", but I can't exclude these things have side effects that lead to crashes.

src/renderers/imgui_dx11/buffers.rs
```diff
@@ -22,26 +20,35 @@ struct VertexConstantBuffer([[f32; 4]; 4]);
...
                 .CreateBuffer(
                     &D3D11_BUFFER_DESC {
                         ByteWidth: std::mem::size_of::<VertexConstantBuffer>() as u32,
                         Usage: D3D11_USAGE_DYNAMIC,
-                        BindFlags: D3D11_BIND_CONSTANT_BUFFER.0,
-                        CPUAccessFlags: D3D11_CPU_ACCESS_WRITE.0,
-                        MiscFlags: 0,
+                        BindFlags: D3D11_BIND_CONSTANT_BUFFER,
+                        CPUAccessFlags: D3D11_CPU_ACCESS_WRITE,
+                        MiscFlags: D3D11_RESOURCE_MISC_FLAG(0),
```
veeenu, line 33: I wish that for all of these instances we could just put a big fat `0` in there and have the type system figure it out via the plethora of conversion traits -- `From`, `Deref`, whatever -- which is one of Rust's strong suits.

src/renderers/imgui_dx11/device_and_swapchain.rs
```diff
@@ -79,28 +77,28 @@ impl DeviceAndSwapChain {
...
             let p_back_buffer: ID3D11Resource = swap_chain.GetBuffer(0).expect("GetBuffer");
+            let mut back_buffer = None;
 
-            let back_buffer = dev
-                .CreateRenderTargetView(&p_back_buffer, null_mut())
+            dev.CreateRenderTargetView(&p_back_buffer, None, Some(&mut back_buffer))
                 .expect("CreateRenderTargetView");
 
-            dev_ctx.OMSetRenderTargets(&[Some(back_buffer.clone())], None);
+            dev_ctx.OMSetRenderTargets(Some(&[Some(back_buffer.clone().unwrap())]), None);
```
veeenu, line 85: If Hawkwood was a Rust developer, I'm sure having to do this would give him conniptions.

src/renderers/imgui_dx11/shader_program.rs
```diff
@@ -153,105 +162,124 @@ impl ShaderProgram {
...
-            rasterizer_state,
-        })
+        Ok(shader_program)
     }
 
     pub(crate) unsafe fn set_state(&self, dasc: &DeviceAndSwapChain) {
-        dasc.dev_ctx().VSSetShader(&self.vtx_shader, &[]);
-        dasc.dev_ctx().PSSetShader(&self.pix_shader, &[]);
+        dasc.dev_ctx().VSSetShader(&self.vtx_shader, Some(&[]));
+        dasc.dev_ctx().PSSetShader(&self.pix_shader, Some(&[]));
```
veeenu, line 279: In this part specifically I have _no idea_ what passing in `Some(&[])` vs `None` does, and can't find any documentation.

Godnoken:
> I have consciously decided to stick with `0.39` for `hudhook` because over the last few releases `windows-rs` has gotten _way way worse_ in terms of idiomatics and ergonomics, while no new APIs were added.

I definitely noticed a lot of icks with it, especially that it seems very inconsistent overall. Although it always seemed to be. Will most likely take quite some time before it gets better.

> The day `windows-rs` becomes again an actual idiomatic mapping around the Windows API and appropriately leverages the RAII patterns and pointer-like structures, instead of adding counterintuitive undocumented breaking changes, I'll resume with keeping its updates aligned with `hudhook`, but until then, I've decided not to do that because it would block 100% of my work on the dependents of `hudhook`.

I completely understand & accept that. I'll figure it out on my end either way.

> I honestly feel bad saying all of this because I'm sure `windows-rs` is super hard to manage, but unfortunately lately I don't see a way to rely on it that doesn't cause me huge troubles.

Yup, I have the outmost respect for the work that is being done. Right now it seems like Kenny is the one doing the majority of the work alone, and it have got to be extremely difficult to handle all of it.

> If you need DX12 stuff, using `windows-sys` directly would probably be your best bet as a pointer in there is a pointer in the MS docs; whereas in `windows-rs` currently `Option<T>` might just as well be `T*` in a module`T**` in another, and a straight `T` in yet another, with zero consistency and little relationship to the docs.

Yep, might look into it. Not sure how it would play in though since I need to pass the devices from hudhook to my code and then use it. Hence the need for a bump. Don't know if there is an easier way to work between versions.

src/renderers/imgui_dx11/shader_program.rs
Godnoken: Me neither, I first tried passing just None & then changed it back to &[] because of the crashes, hoping that things would fall back in to place magically lol.

Godnoken:
Thank you for the quick response - you can close this now if you like.

vSylva:
I also upgraded local hudhook's windows dependency to 0.48 and it seems to work fine on dx12. However, this can be risky for open source libraries. We can create a new branch to make some changes, which is suitable for some libraries using windows-0.48

veeenu:
Just noticed that there is a new release apparently, `0.49`. I will try and bump to that, see if it's a little better.

src/renderers/imgui_dx11/shader_program.rs
vSylva: From ChatGPT


VSSetShader is a function in DirectX used to set the vertex shader. It has two parameters:

1. pVertexShader: a pointer to the vertex shader. This parameter is required as it tells DirectX which vertex shader to use to process vertex data.

2. pClassInstances: a pointer to an array of D3D11_CLASS_INSTANCE structures. This parameter is optional and allows you to dynamically create vertex shader instances at runtime. If you don't need to create instances, you can set this parameter to NULL.

Other related parameters, such as shader constant buffers and input layout, need to be set before setting the vertex shader. The VSSetShader function simply binds the specified vertex shader to the vertex shader stage of the rendering pipeline.

It should be noted that this function is only applicable to DirectX 11 and higher versions. In earlier versions of DirectX, setting the vertex shader requires the use of different functions.

Godnoken: Hehe, we're talking about the window-rs crate specifically. If only it was so easy to look at MDSN at all times..

Godnoken:
> Just noticed that there is a new release apparently, `0.49`. I will try and bump to that, see if it's a little better.

Yeah, I saw that when I started the 0.48.0 bump, but for whatever weird reason I couldn't compile with 0.49.0. Maybe because it isn't bumped on crates.io yet?

I solved all my DirectX issues by the way - but IMO it doesn't hurt to keep up if you've got the time.

src/renderers/imgui_dx11/shader_program.rs
vSylva: Hehe. I remember the parameter can be set to None

veeenu:
0.51.x is out -- I'll probably try bumping to that right before releasing 0.5.0, see if it fixed anything, though I'm still not super confident about the potential outcome. Closing this for the time being.

PR #99, opened 2023-05-20
Fix #[warn(unused_mut)]

vSylva:


veeenu (approved):
Thank you for this! 🚀

PR #104, opened 2023-05-27
Add raw input handling

veeenu:
This PR adds raw input handling to the window procedure.

The code was shamelessly stolen from #94, and refactored a bit.

Question for @Godnoken: I tested this against Dark Souls III and Elden Ring; no changes in behavior were detected. In your branch, what is the purpose of adding the handling of raw input? Are there cases in which only raw input or only "normal" input messages (e.g. `WM_SYSKEYDOWN` etc) are sent to the wnd proc?


src/hooks/common/raw_input.rs
```diff
@@ -0,0 +1,176 @@
...
+        }
+        if flags & RI_KEY_E1 != 0 {
+            code |= 0xe100;
+        }
+        code
+    };
+
+    // Check the key status.
+    let is_key_down = flags == RI_KEY_MAKE;
+    let is_key_up = flags & RI_KEY_BREAK != 0;
```
veeenu, line 114: In general, I want to start pushing this repo in the direction in which everything is organized and commented in this way. I've been slacking off on that, but I realized that this way the understandability of a complex codebase skyrockets.

Godnoken:
> The code was shamelessly stolen from https://github.com/veeenu/hudhook/pull/94, and refactored a bit.

And mine is shamelessly stolen from Reshade's C++ codebase. Even though the code that you have now is quite dramatically different from my PR, I believe a notice of that would be in order before a merge down the road.

> Question for @Godnoken: I tested this against Dark Souls III and Elden Ring; no changes in behavior were detected. In your branch, what is the purpose of adding the handling of raw input? Are there cases in which only raw input or only "normal" input messages (e.g. `WM_SYSKEYDOWN` etc) are sent to the wnd proc?

Yes, plenty of games only use raw input, and that is hence the only thing that will be sent to the ``wndproc``. Oblivion is one that I can think of just off the bat.

---

The code looks far more readable than mine and I would have liked to update ``io`` directly like we already do, however my concern that I voiced earlier was the fact that I don't think that is something that would work when using hooks.
We'd end up having 3 possible hooks that could be locking the renderer at the same time to get access to the ``io`` and thus end up deadlocking (..right?) or becoming a massive overhead. At least as far as my knowledge goes.. 😄 

Of course this is something that can be solved down the road when and if you decide that it is an option that you want to implement.

Another concern (unrelated to adding raw input) that exists since before is the random input block happening in some games. Not sure if you ever got them yourself. I've acquired a decent chunk of experience since then and off the top of my head I bet it is happening because we're using ``try_lock()`` on the renderer to be able to run ``wndproc`` and thus we can't register input sometimes. Definitely something to think over.

veeenu:
> And mine is shamelessly stolen from Reshade's C++ codebase. Even though the code that you have now is quite dramatically different from my PR, I believe a notice of Crosire's work would be in order before a merge down the road.

Good point!
 
> Yes, plenty of games only use raw input, and that is hence the only thing that will be sent to the `wndproc`. Oblivion is one that I can think of just off the bat.

So the current hudhook `main` wouldn't work with Oblivion? Good to know -- means this addition alone should already cover a lot more use cases. I wonder if this could ever result in double inputs in certain cases? Probably nothing to worry about until we run actually run into such an issue, but if that's the case then we'll have to find a way of detecting those cases and only processing either raw or normal input.

> The code looks far more readable than mine and I would have liked to update `io` directly like we already do, however my concern that I voiced earlier was the fact that I don't think that is something that would work when using hooks. We'd end up having 3 possible hooks that could be locking the renderer at the same time to get access to the `io` and thus end up deadlocking (..right?) or becoming a massive overhead. At least as far as my knowledge goes.. 😄

That sounds right, yeah. A possible solution that comes to mind is using `mpsc` inside of the window procedure and just shuttling off the job of updating the imgui context to another thread, where the mutex is held very briefly only when a new message comes in. This should result in almost no overhead. 

> Another concern (unrelated to adding raw input) that exists since before is the random input block happening in some games. Not sure if you ever got them yourself. I've acquired a decent chunk of experience since then and off the top of my head I bet it is happening because we're using `try_lock()` on the renderer to be able to run `wndproc` and thus we can't register input sometimes. Definitely something to think over.

Yeah, big time -- I feel like I was a bit too heavy handed with mutexes 😅 I want to investigate whether a different synchronization mechanism could still work. There are some things that could be made immutable, like the stored `wnd_proc` to be called back; we know that that procedure is going to stay the same as long as we're hooked so there's no need to synchronize it. Other fields are probably similar in nature. This is a very big undertaking but it's probably worth looking into it.

PR #108, opened 2023-06-14
Remove local minhook

vSylva:
Perhaps this is a better way

Replace local minhook with minhook_raw = "0.1.3"

https://crates.io/crates/minhook_raw


veeenu:
I would prefer to keep `minhook` vendored locally.

The reasoning is that `minhook` has not received any updates in quite a long time, is very stable, and is very small; the same cannot be said of individual GitHub repositories. 

In fact, at first we used to just have `build.rs` here and `minhook` as a submodule, and later came the decision of vendoring it altogether to minimize risks.

That being said, I think that a crate that exposes FFI bindings for `minhook` is definitely a good idea, as not everyone wants to go through the pain of maintaining a `build.rs` themselves.

vSylva:
> I would prefer to keep `minhook` vendored locally.
> 
> The reasoning is that `minhook` has not received any updates in quite a long time, is very stable, and is very small; the same cannot be said of individual GitHub repositories.
> 
> In fact, at first we used to just have `build.rs` here and `minhook` as a submodule, and later came the decision of vendoring it altogether to minimize risks.
> 
> That being said, I think that a crate that exposes FFI bindings for `minhook` is definitely a good idea, as not everyone wants to go through the pain of maintaining a `build.rs` themselves.

I think it's better to separate minhook out, but your viewpoint may be more appropriate😄

veeenu:
I don't think either solution is inherently superior. In general I think using a separate crate would be more appropriate in most situations.

For the specific case of `hudhook`, though, I think the benefits of having it vendored (no risk of third party maintenance/availability issues, guaranteed stability over time) outweigh the costs (extra maintenance burden).

Jakobzs:
> > Btw -- I just remembered that @Jakobzs already createad a [minhook](https://crates.io/crates/minhook) crate a while ago! 😄
> 
> I saw this crate
> 
> This crate does not expose the original functions

Bit late to reply here, but that is true. It is more a "wrapper" around it, although API wise I believe I expose all possible API that can be called, and also guarantee initialization etc (so the user does not have to worry about it). While it is not too much different from using the raw API, I feel it does provide a little quality of life.

Issue #96, opened 2023-05-08
Bug: D3D12CreateDevice on D3D11 Hook


vars1ty:
Hello, I just recently found out about this project and it seems to suit all of my needs, one "small" issue though is that when I try and inject the DirectX 11 Example provided (built with `x86_64-pc-windows-gnu`), it errors with the message: `Force load module failed:failed finding address of d3d1 D3D12CreateDevice`, I'm using Cheat Engine to inject it.

I also tried to use the OpenGL 3 example and it gives the same output. I'm on Linux but I have also tried to build and run it on a Windows VM and got the exact same result.

As far as I can tell, D3D12CreateDevice should only be called for the DirectX 12 hook, so I'm completely lost as to why it's crashing on it for hooks that aren't related to it (?).
I tried to use another DirectX 11 Hook, which is this: https://github.com/ynuwenhof/imferris and it worked just fine, both on Linux and on Windows.

Are there any steps I can take to potentially solve this issue? Thanks!

veeenu:
Hey, thank you for reporting this!

I honestly have never tried `hudhook` on Linux and I'm not sure if there are translation layers, but this definitely should not happen on the Windows VM.

You're right, `D3D12CreateDevice` should not be on the code path at all for anything but the Dx12 hook.

Furthermore, I searched `failed finding address` in this repo and I don't get exact results -- I'm not sure that error comes from anything related to `hudhook`, maybe it has something to do with how it's structured.

Could you try taking a look at the [tutorial](https://veeenu.github.io/hudhook/)? It should help you build a small minimal example pretty quickly, dll injector included. Do you get the same errors if you inject via that?

vars1ty:
Hello! I'll test it in the coming days, and the error code is retrieved from Cheat Engine, I should of been more clear on that part.

veeenu:
I tried looking into Cheat Engine's code base but I'm not familiar enough, and could not find an explanation in the limited time I manage to allocate to the task. But yeah if it is an issue with Cheat Engine, using an external injector that uses other techniques (such as the wrappers provided by `hudhook` itself) might be able to help.

PR #116, opened 2023-06-29
Lock hooks and injection behind feature flags

vars1ty:
Locked all hooks and the injection part behind the feature flags, injection is behind `inject`.

Regarding #96 - It's fixed with this as I only enable OGL3, but there's another new issue that only occur with the Linux -> Windows build process, where it fails get some user profile directory (logged by CE).
Not a big deal, but if you got time to look into it, please do.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,0 +1,2 @@
+[build]
+target = "x86_64-pc-windows-gnu"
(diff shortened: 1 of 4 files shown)


veeenu:
Oh my, that was an oversight and a half on my end 😅 

Thank you for this PR! LGTM 😄 

Could you please elaborate on your issue (on #96, or a new issue, should be good) so I can try and reproduce it?

PR #126, opened 2023-08-19
Change top-level API

veeenu:
This PR reorganizes the top-level module and results in a simpler user-facing API.

This is achieved by changing the `Hooks` trait so that its implementors must no longer be in charge of applying the hooks themselves, but must implement a function that returns a slice of `MhHook`s which will be acted upon by the library itself.

A builder pattern API was added that allows initializing potentially multiple different hooks via repeated invocations of the `with` method.

The entry point generator macro's API was not changed, but its generated code was greatly simplified, and the `lifecycle` and `lifecycle::global_state` modules were removed as those implementation details are now hidden behind the new `HudhookBuilder` object.

This has been tested on the Dark Souls III practice tool, and everything seems to be working just fine.

@Godnoken this could be of particular interest to you, as this now allows for the addition of optional hooks and we may now implement the input blocker properly, in a way that allows it to be opt-in. Do you mind giving this a spin before I merge it in?


veeenu:
Merging this in anyway to unblock #127.

PR #127, opened 2023-08-22
Bump `windows` to 0.51

veeenu:
This PR bumps `windows-rs` to the latest 0.51 version.

There are some new patterns to deal with optional out params that `windows` isn't handling in a very elegant way, e.g. options of options of mutable pointers where the difference between `None` and `Some(null_mut())` is not particularly evident, so some helper functions to deal with these cases should be added and applied:

- `out_param` for pre-allocated out parameters
- `try_out_param` for pre-allocated out parameters in a fallible function
- `out_ptr` for out pointers allocated by the function itself
- `try_out_ptr` for out pointers allocated by a fallible function

Example of usage:

```rust
let desc = try_out_param(|&mut out| device_and_swap_chain.GetDesc(out))?;
```

Blocked by #126

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -78,9 +78,10 @@ crate-type = ["bin"]
 imgui = "0.11"
 imgui-opengl = "0.1"
 parking_lot = "0.12"
-windows = { version = "0.39.0", features = [
+windows = { version = "0.51.0", features = [
   "Win32_Devices_HumanInterfaceDevice",
   "Win32_Foundation",
+  "Foundation_Numerics",
   "Win32_Security",
   "Win32_System_Console",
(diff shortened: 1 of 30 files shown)


veeenu (commented):
While the new conventions in `windows-rs` are a bit verbose, they do feel more consistent, and having to repetitively instantiate `None` mutable options for storing pointers is easily (if not too elegantly) dealt with by using a couple simple helper functions.

Overall I'm very happy of this version bump because the code feels cleaner in the right spots and we can finally move on and stay aligned with new versions going forward.

As a mostly unrelated bonus, with this PR we officially compile on stable Rust and can finally drop the `nightly` requirement 🎉

PR #136, opened 2023-11-01
Fix OpenGL3 comment in README

Jakobzs:
Biggest PR ever, very crucial fix


veeenu:
I mean, the optics of a wrong README aren't great, so I wouldn't devalue it 😁

PR #140, opened 2023-12-02
Fixed hooking comment

jacobtread:
## Description

Fixed comment referring to DirectX 9 instead of OpenGL 3 

## Changes
- Updated comment in lib.rs


veeenu:
Thank you!

Issue #137, opened 2023-11-03
DX9 ImGui unresponsive


vars1ty:
The DirectX9 ImGui doesn't react nor detect proper mouse position in the same game as in #132.
It logs the position as "100, 100" and never recognizes any change, apart from sudden 1-frame drops to like 0, 0.

Code is the sample code from examples.

veeenu:
I can't work on this at the moment, but could you try deleting the first `#[cfg]` statement and the entirety of the second block here?

https://github.com/veeenu/hudhook/blob/bebcddcb82f462061c1101383001677b4af276ee/src/hooks/dx9.rs#L39-L51

So that only this part remains:

https://github.com/veeenu/hudhook/blob/bebcddcb82f462061c1101383001677b4af276ee/src/hooks/dx9.rs#L40-L44

Let me know if that fixes things; in that case, please open a PR so I can merge that in.

vars1ty:
Hi, sorry for the late response, been busy.
I tried that just now, and it makes no difference from what I can tell.
The game in question is an offline RPG which does some **very** weird hacky things to its window, here it is if you'd like to take a look when you've got the time: https://archive.org/download/star-stable-win-iso-en.-cz.-fi.-hu.-de.-no.-se_202011/Star_Stable_Win_ISO_EN.CZ.FI.HU.DE.NO.SE.zip

Godnoken:
Unfortunately, I can confirm this issue is still present in Fallout 3.

veeenu:
Haven't found much use for raw input personally, so I can't attest firsthand to the quality of the implementation... I'd gladly accept any fixes for it though.

Godnoken:
I was wrong. It's not related to raw input. The Elder Scrolls and Fallout series uses DirectInput. The reason why the mouse pos is no longer updated at all is because of the fact that we no longer update the io.mouse_pos inside the render function. However, by doing so, we're right back to square one with the mouse pos alignment issues mentioned in https://github.com/veeenu/hudhook/discussions/33#discussioncomment-3943592

It seems to me that the only working solution is hooking into DirectInput unfortunately.

veeenu:
Well, if necessary, `mouse_pos` can be updated [here](https://veeenu.github.io/hudhook/rustdoc/hudhook/trait.ImguiRenderLoop.html#method.before_render) with whichever source of data. Honestly I think we should offer additional hooks as opt-in utilities, with DirectInput being one of them.

Issue #107, opened 2023-06-08
Allow DirectX hooking using existing device pointer


Earthnuker:
Hi,

would it be possible to add a function to initialize hudhook with an existing Direct9Device pointer? one of my projects is an overlay for a DirectX8 game which i'm running through [d3d8to9](https://github.com/crosire/d3d8to9) but initialize hudhook using the following code:
```rust
// imports, Overlay struct definition and empty ImguiRenderLoop impl left out for brevity
static OVERLAY_INITIALIZED: AtomicBool = AtomicBool::new(false);
pub fn init() -> Result<()> {
    if OVERLAY_INITIALIZED.swap(true, Ordering::SeqCst) {
        bail!("Tried to initialize overlay twice!");
    }
    // Passing a handle to the module around into the function would be a bit cumbersome, hence GetModuleHandleA()
    let hmodule = unsafe {std::mem::transmute(windows::Win32::System::LibraryLoader::GetModuleHandleA(None)?)};
    hudhook::lifecycle::global_state::set_module(hmodule);
    let hooks: Box<dyn hooks::Hooks> = Overlay::new().into_hook::<ImguiDx9Hooks>();
    unsafe {hooks.hook()};
    hudhook::lifecycle::global_state::set_hooks(hooks);
    println!("OK!");
    Ok(())
}
```
fails with the following error message:
```
panicked at 'IDirect3DDevice9::CreateDevice: failed to create device: Error { code: 0x8876086C, message:  }', C:\Users\Earthnuker\scoop\persist\rustup-msvc\.cargo\registry\src\index.crates.io-6f17d22bba15001f\hudhook-0.4.0\src\hooks\dx9.rs:324:6
```

veeenu:
Hi!

This would probably require a big change in the API, though I admit if we had access to devices _before_ having the hooks we could skip the whole song and dance of initializing dummy devices for finding out the `Present` address. Which would be pretty good.

That panic doesn't depend on the device that's found in the game, though; rather in a failure to create a new dummy device with DirectX 9.

I don't know about the inner workings of `d3d8to9` but it feels like it could be one of these two:
- under the hood the game does Dx8 calls that get routed to Dx9 calls (and hence a DirectX 9 device is available somewhere)
- the game has Dx9 calls available that actually get routed to Dx8 internally (and hence we would need a DirectX 8 specific hook).

[...]

Issue #103, opened 2023-05-24
[DX9] Some games do not reset at window resizing


Godnoken:
- Warhammer 40K: Dawn of War;  Displays ImGui fine if injected on launch, but if you change resolution in-game then ImGui disappears due to no trigger of a device reset.
 - Heroes of Might & Magic 5;  If injected on launch, ImGui doesn't show up & resolution change doesn't trigger device reset

The solution could be to save the previous size & check if it changes on each frame. Since we already call `get_window_rect`, hopefully that is enough, but it could very well be so that the device's viewport needs to be checked as well like in the OpenGL solution.

Since this is a fairly isolated thing, I'm wondering  - Is the overhead of checking the viewport basically non-existent or is it something that has to be considered?

----

There is also another game (don't remember which one) with a similar issue. Might even have been a DirectDraw game that I ran with [dxwrapper](https://github.com/elishacloud/dxwrapper), so I'm not too sure of the relevancy here since it may be so that there is no equivalent to `Reset` in those games.

The only solution I could come up with was to read for calls to `WM_ACTIVATEAPP` and subsequently reset the device manually from there.

```
 WM_ACTIVATEAPP => {
            if let Some(device) = D3D9DEVICE.get() {
                let mut present_params = D3DPRESENT_PARAMETERS {
                    Windowed: BOOL(1),
                    SwapEffect: D3DSWAPEFFECT_DISCARD,
                    BackBufferFormat: D3DFMT_UNKNOWN,
                    ..core::mem::zeroed()
                };
                device.Reset(&mut present_params).unwrap();
            }
        }
```

It's not the prettiest solution & would require us to expose the device somehow like talked about in https://github.com/veeenu/hudhook/issues/80

Godnoken:
#158 does fix the Warhammer 40K issue. Heroes Of Might And Magic 5 still have issues present. Injecting early doesn't render imgui (alt-tabbing and tabbing back make it pop up). Changing the resolution does now keep imgui window rendering, however, the game starts running at like 0.5fps, lol. This issue exists on older hudhook versions as well.

Issue #79, opened 2023-02-24
Game input not registering correctly


Godnoken:
There's an issue with game input that pops up in some games, seemingly primarily DX9 games, although I have had limited possibilities to try lots of games with the other renderers.
Basically, input sometimes registers fine, and other times not.

The "WndProc called before hook was set" log from the imgui_wnd_proc function always show up in the games where this happens.
I'm not sure what the cause of this is yet. I'm going to investigate further and I'd be happy if anyone knows anything that'd help.

Godnoken:
Is this something you noticed when you built the DX9 renderer @FrankvdStam?
I saw that you mentioned double rendering issues here https://github.com/veeenu/hudhook/pull/21/commits/4241d0b62e987ba2ec9093c6dd2802e8e4632b5c. Is that something you completely fixed, and if not, could it be related to this..?

FrankvdStam:
Double rending was fixed. Only issue I know of that I didn't fix has to do with scaling/mouse positions. What you're having seems more like a race condition, where the WndProc hook is not always installed at the right time. My guess would be to inspect the initialization steps carefully, made add some log steps to figure out if something is crashing and burning, causing the hook to not be installed sometimes.

Godnoken:
Hmm, okay, interesting. I will start there. However, just to be clear - Game input isn't completely dead, it's just that sometimes input doesn't reach the game, and these messages always log in the same games.
But as said previously, this is not something unique for the DX9 renderer, I have at least one case of it happening in DX11 with the game Niche. It just seems (as far as I've tested) that it is more prominent with DX9 games, and it feels likely that this is something that could exist in all three of Hudhooks DirectX renderers.

I'm starting to understand how all of this works but I'm still miles from being able to make good educated guesses, so really - any input is super valuable!

Godnoken:
Oh. Here's an interesting one.. Disabling Steam overlay fixes the game input on Shelter2 & Car Mechanic Simulator 2015 (both DX9), while Niche (DX11) still breaks.
It seems that Steam overlay creates its own device in some games and interferes with our DX9 device.. Possibly we could check the return address and only call if it is our device..? Not sure if that makes sense.

veeenu:
[...]

PR #143, opened 2024-01-26
Unified renderer

### LINKED ISSUE #111: Build `wgpu` renderer
`wgpu` is a crate that supports multiple [backends](https://docs.rs/wgpu/latest/wgpu/enum.Backend.html).

This would currently leave out DirectX 9, but I wonder if we could leverage it to merge the DirectX11, DirectX12, OpenGL renderers and also add instant support for Vulkan.

### LINKED ISSUE #142: Incompatibility Issue with Sea of Thieves (DX11)
I am currently experiencing issues while trying to implement the HudHook library into the Sea of Thieves game. Despite following all the instructions provided in the documentation, I encountered several problems that are preventing me from using it smoothly.

Firstly, when utilizing the master branch of the HudHook library, I experience random freezing of the game and I get "Could not lock in WndProc" every ~10sec in the trace console. Occasionally, the game runs without any apparent issues; however, the in-game menu fails to show up. Moreover, I encounter rendering glitches similar to those demonstrated in this video: [Rendering Glitches Video](https://www.youtube.com/watch?v=ZR8cGKQfjVQ)

Secondly, as an alternative approach, I attempted to use the unified-renderer branch instead. While this version resolves the previous rendering concerns, there is now another issue where neither the mouse nor the keyboard functions properly within the ImGUI window. Consequently, controlling or interacting with the elements inside the ImGUI window becomes impossible. To provide further clarification regarding these challenges, please refer to this additional video demonstration: [Demonstration Video](https://www.youtube.com/watch?v=TafFp3wb3qA)

Would you happen to have any suggestions for addressing this issue?I have endeavored to modify your lib to try to fix this issue but without success. 

Despite these hurdles, I genuinely appreciate the efforts invested in developing such a fantastic lib.

Thank you in advance for your time and support.

veeenu:
This is a **MASSIVE** refactor that completely changes the underlying rendering engine, which is now a unified DirectX12 renderer that is then composited over games' windows via [DirectComposition](https://learn.microsoft.com/en-us/windows/win32/directcomp/directcomposition-portal), with barely any public API change.

It seamlessly works with Dark Souls III and Elden Ring already, and the integration test cases.

Being such a breaking change, I will shamelessly ping all people that may be affected as I don't have a lot of control over corner cases but really don't want to merge this in before I'm sure I'm not disrupting anything. I would be super grateful if you'd be up for checking this out. I expect some of you may have moved on to other projects, in that case I hope you won't be annoyed by the ping!

## Explanation

This PR implements a DirectX 12 renderer in `hudhook::renderer::dx12` which renders over an off-screen surface, and then uses DirectComposition to composite the off-screen surface on top of the HWND, completely sidestepping the games' native rendering engine which will run on its own.

`Present` hooks now only serve the purpose of actually triggering a render and a composition and the device objects aren't touched at all. In fact, most of the hooks' rendering logic except building the actual hooked function implementations is now shared and identical across all hooks.

As a result, hooks are now tiny compared to what they were before and their maintenance is going to be very easy. Similarly, with a single renderer there is much less code to maintain.

Additionally, this will make it _extremely_ simple to add new renderers 

## Affected issues

As a bonus, this PR _should_ fix the following issues, or at the very least give us a cleaner path to their resolution.

If you are getting pinged, it would mean the world if you could give this branch a spin and report back if it fixed the associated issue.

- #79
  @Godnoken @camas since the renderer does not lock DirectX9 specific resources, it could alleviate/solve the issue. (need to investigate)
- #94
  @Godnoken I have found a different way of blocking input that doesn't involve rerouting many input functions: just use a transparent overlay window. It's not implemented but I think we could explore that avenue and it could work for all hooks.
- #103 
  @Godnoken this branch uses `WM_SIZE` for resize events which should sidestep everything that causes your issue altogether. (might close the issue)
- #107 
  @Earthnuker @SkyLeite this greatly reduces the issue surface; now that the hook code is drastically simpler, we could try a number of solutions, provided this one somehow doesn't solve the issue already. (need to investigate)
- Closes #111.
  We won't need a renderer with multiple backends after all!
- #137 
  @vars1ty not sure about the specific game, but maybe things may have changed enough that this now works correctly. (need to investigate)
- Closes #142 
  @Thoxy67 this branch should most definitely work for you now.

Thank you all for your patience and sorry again for the ping! 😅


src/hooks/opengl3.rs
```diff
@@ -1,246 +1,44 @@
...
-            let mut imgui_renderer = ImguiRenderer {
-                ctx: context,
-                renderer,
-                wnd_proc,
-                game_hwnd: hwnd,
-                resolution_and_rect: None,
-            };
-
-            // Initialize window events on the imgui renderer
-            ImguiWindowsEventHandler::setup_io(&mut imgui_renderer);
```
veeenu, line 73: Note to self: this part of logic is missing and should be added

camas:
Unfortunately #79 still an issue. Tested with 7791c49

Fails soon after injection when it can't lock the render engine, with a bunch of WndProc lock fails before.

```
19:00:12 [DEBUG] (1) hudhook::hooks: WNDCLASSEXW { cbSize: 80, style: WNDCLASS_STYLES(3), cbClsExtra: 0, cbWndExtra: 0, hInstance: HINSTANCE(140696328601600), hIcon: HICON(0), hCursor: HCURSOR(0), hbrBackground: HBRUSH(0), lpszMenuName: PCWSTR(0x0), lpszClassName: PCWSTR(0x7ffbf3eff8e8), hIconSm: HICON(0) }
19:00:12 [DEBUG] (1) hudhook::hooks: HWND(788424)
19:00:12 [INFO] Output window: IDXGISwapChain(IUnknown(0x22b60745720))
19:00:12 [INFO] Desc: DXGI_SWAP_CHAIN_DESC { BufferDesc: DXGI_MODE_DESC { Width: 1920, Height: 1080, RefreshRate: DXGI_RATIONAL { Numerator: 0, Denominator: 1 }, Format: DXGI_FORMAT(28), ScanlineOrdering: DXGI_MODE_SCANLINE_ORDER(0), Scaling: DXGI_MODE_SCALING(0) }, SampleDesc: DXGI_SAMPLE_DESC { Count: 1, Quality: 0 }, BufferUsage: DXGI_USAGE(48), BufferCount: 2, OutputWindow: HWND(1771696), Windowed: BOOL(1), SwapEffect: DXGI_SWAP_EFFECT(3), Flags: 2114 }
19:00:25 [DEBUG] (2) hudhook::hooks::render: Could not lock in WndProc
...
19:01:01 [DEBUG] (2) hudhook::hooks::render: Could not lock in WndProc
19:01:01 [ERROR] Could not lock render engine
```

veeenu:
@camas I stalked your GitHub and noticed you are using `hudhook` for Golf With Your Friends. I have that game so I can take a look at it directly! But yeah I can easily see how that's a synchronization issue. It could be solved by using threads and channels but I'm not sure I want the added overhead of those.

Weirdly enough, there is only 3 places now where the render engine is locked that could create contention:
- [here during `render`](https://github.com/veeenu/hudhook/blob/51068020db8e9304be73de171678478cdbbda4b7/src/hooks/render.rs#L77)
- [here during wnd proc](https://github.com/veeenu/hudhook/blob/51068020db8e9304be73de171678478cdbbda4b7/src/hooks/render.rs#L124)
- [here during `resize`](https://github.com/veeenu/hudhook/blob/51068020db8e9304be73de171678478cdbbda4b7/src/hooks/render.rs#L93)

My hunch there is that the wnd proc calls itself recursively, locks once and is not ever able to exit and do anything else. Or the `WND_PROC` static is initialized twice (which shouldn't happen as we use `get_or_init` that has the guarantee of uniqueness) and thus called recursively via `CallWindowProcW`.

veeenu:
I have gotten some feedback and will now merge this in as there are no apparent regressions for the time being.

vars1ty:
Thanks for the ping, feel free to ping me whenever as I'm more than happy to test out features whenever I can.

I'll update to the latest commit, fix a few errors due to the changes and then test it in both OpenGL and then later DirectX9, will let you know how it goes 👍🏽

veeenu:
Thank you 🙏

PR #145, opened 2024-01-28
Add before_render function in ImguiRenderLoop

Thoxy67:
Like I mention [here](https://github.com/veeenu/hudhook/issues/142#issuecomment-1913642965) this is the modification that work after some test to force show the cursor and separate the key list from the `setup_io`

If I place the "show cursor" code directly inside the loop without proper synchronization or toggling (i.e., changing based on key pressed once), it results in crashes after some time.


Thoxy67:
I have added this config options : 

```rust
// Here are the default values
pub static mut SHOW_UI: bool = true; // Show the UI by default when injected and also at runtime
pub static mut FORCE_SHOW_CURSOR_ON_UI_SHOW: bool = false; // Force cursor to show when ui show

// Here are defined the default keycode (keycode as u16, None = disabled function)
pub static mut SHOW_UI_KEY: Option<u16> = None; // Define the keycode for hide / show the ui
pub static mut SHOW_CURSOR_KEY: Option<u16> = None; // Define the keycode for hide / show the cursor
```

I have also updated this libs : 

```toml
[dependencies]
once_cell = "1.19.0"
windows = "0.52.0"

[build-dependencies]
cc = "1.0.83"
```

src/renderer/dx12.rs
```diff
@@ -511,18 +536,18 @@ impl RenderEngine {
...
-        let mut ctx = ctx.borrow_mut();
-        let ui = ctx.frame();
-        render_loop(ui);
-        let draw_data = ctx.render();
-
-        if let Err(e) = unsafe { self.render_draw_data(draw_data, idx) } {
-            eprintln!("{}", e);
-        };
-
+        if unsafe { SHOW_UI } {
```
veeenu, line 539: This is a kind of usage of `unsafe` that should be avoided **at all costs**.

We are operating in a concurrent environment, and statements like this **will** lead the compiler to take some unpredictable optimizations that lead to undefined behavior.

src/renderer/dx12.rs
```diff
@@ -592,28 +617,15 @@ impl RenderEngine {
...
-        io[Key::Escape] = VK_ESCAPE.0 as _;
-        io[Key::A] = VK_A.0 as _;
-        io[Key::C] = VK_C.0 as _;
-        io[Key::V] = VK_V.0 as _;
-        io[Key::X] = VK_X.0 as _;
-        io[Key::Y] = VK_Y.0 as _;
-        io[Key::Z] = VK_Z.0 as _;
+        for i in keys::KEYS {
+            io[i.0] = i.1 .0 as _;
+        }
```
veeenu, line 625: This is the part that we can keep, and it's a very good idea.

src/renderer/dx12.rs
```diff
@@ -592,28 +617,15 @@ impl RenderEngine {
...
+
+        // Add to key_press static variable to store the status for toggle
+        unsafe { KEY_PRESS = io.keys_down };
```
veeenu, line 628: Same comment about `unsafe` as above. I recommend the excellent book ["Rust Atomics and Locks"](https://marabos.nl/atomics/) (web version available for free IIRC) to better understand the implications of this.

src/renderer/dx12.rs
```diff
@@ -1071,12 +1083,12 @@ impl RenderEngine {
...
-                (t + b) / (b - t),
-                0.5,
-                1.0,
-            ]]
+            [
+                [2. / (r - l), 0., 0., 0.],
+                [0., 2. / (t - b), 0., 0.],
+                [0., 0., 0.5, 0.],
+                [(r + l) / (l - r), (t + b) / (b - t), 0.5, 1.0],
+            ]
```
veeenu, line 1091: This should be formatted in the original way via `cargo +nightly fmt --all`.

src/renderer/keys.rs
veeenu, line 1: This is great 👍

veeenu (changes requested):

I can accept the part about the `Key` enum setup, but the rest of the solution with the global statics really doesn't belong in the library and should be full responsibility of the client code, and would introduce undefined behavior.

As an example, check my solution for the [Elden Ring practice tool](https://github.com/veeenu/eldenring-practice-tool/blob/528aab05c2e256b0c5c0be9e32e1fc7f306c1e1c/practice-tool/src/lib.rs#L430-L435): there is an internal game method that enables or disables the rendering of a cursor, and the logic for flipping that is offloaded to the tool's code without the need of introducing additional abstraction and opinionated flow in the library.

Adding functionality that's disconnected from the main flow of the program makes it hard to maintain, hard to document, and hard to reason about when inspecting bugs. A different solution should be sought.

A possible alternative solution would be to augment the [`ImguiRenderLoop`](https://github.com/veeenu/hudhook/blob/unified-renderer/src/lib.rs#L213) trait with a lifecycle hook that accepts a `&mut Context`:

```rs
pub trait ImguiRenderLoop {
    // ... other methods
    fn before_render(&mut self, _ctx: &mut Context) {}
}
```

This could be called [here](https://github.com/veeenu/hudhook/blob/unified-renderer/src/hooks/render.rs#L82-L84) before rendering and would allow the user to perform their own logic such as setting `mouse_draw_cursor` after querying the key press states. It cannot be done during `render` because creating a `Ui` object shared-borrows the `Context` as well, making it immutable in its lifetime, but we can freely do that in two steps like I'm suggesting.

Anyway I think your use case could be easily solved by using the `ShowCursor` API directly and storing the relevant state (i.e. a boolean "should I show cursor" that's flipped according to button presses) inside of your `ImguiRenderLoop` implementor struct.

Don't take the harsh feedback personally; again, I am very grateful to you for taking the time to contribute!

src/renderer/dx12.rs
Thoxy67: removed

src/renderer/dx12.rs
Thoxy67: Great I think it make the code cleaner

src/renderer/dx12.rs
Thoxy67: removed

src/renderer/dx12.rs
```diff
@@ -304,6 +305,30 @@ pub struct RenderEngine {
...
+                SHOW_UI = !SHOW_UI;
+                if FORCE_SHOW_CURSOR_ON_UI_SHOW {
+                    io.mouse_draw_cursor = SHOW_UI;
+                }
+            }
+        },
+        None => {},
+    }
+}
+
```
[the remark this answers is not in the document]
Thoxy67: removed

src/config/mod.rs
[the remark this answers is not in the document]
Thoxy67: I have removed all the new static config method but i have added the : 

```rust
pub trait ImguiRenderLoop {
    // ... other methods
    fn before_render(&mut self, _ctx: &mut Context) {}
}
```

it seems to work pretty well and pass everything through `ctx.io_mut()...` and the value can be modified here thanks a lotI hope you add this `before_render` feature.

Cargo.toml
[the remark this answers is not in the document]
Thoxy67: For the many test I made this has not break anything but you can try yourself. (tested for SoT and Palworld)

src/renderer/keys.rs
Thoxy67: I'm still learning rust this is nothing but thank you very much i appreciate

src/renderer/dx12.rs
veeenu: I have it explicitly in my editor configuration to use the `+nightly` flag to `cargo fmt` but it comes and goes for me as well for some reason -- I caught this specific one because lately my `rust_analyzer` has been applying the same change automatically over and over with me only noticing the diff in commits.

src/renderer/keys.rs
veeenu: We all have to start somewhere! Hope you enjoy your Rust journey as much as I do! 😄

veeenu:
Thank you for the changes!

Could you please revert the version updates in `Cargo.toml`? We have directly experienced some quirks in the past that resulted in unexpected breakages in `hudhook`, so I'd like to postpone the bumps until I have time to thoroughly go through the changelogs and documentation.

Apart from that, the PR looks good to me now and I plan to merge it in soon (tomorrow morning hopefully) after giving it a manual round of tests.

Thoxy67:
> Thank you for the changes!
> 
> Could you please revert the version updates in `Cargo.toml`? We have directly experienced some quirks in the past that resulted in unexpected breakages in `hudhook`, so I'd like to postpone the bumps until I have time to thoroughly go through the changelogs and documentation.
> 
> Apart from that, the PR looks good to me now and I plan to merge it in soon (tomorrow morning hopefully) after giving it a manual round of tests.

I have reverted the libs versions to old one.

Thanks a lot for everything now I can manage the IO like i wanted to.

Take your time to make your tests and merge wen everything is ok on your side

veeenu:
That's great, thank you!

Could you please run `cargo +nightly fmt --all` so that it passes CI? After that I'll run some tests and we're good to go!

Thoxy67:
> That's great, thank you!
> 
> Could you please run `cargo +nightly fmt --all` so that it passes CI? After that I'll run some tests and we're good to go!

When i call this this change so many things like all this : 

image

I think i need to do it on my linux env this make so many change (actually on my windows dual boot) in the code like updating the crates, change some part of the code in comment and every single c++ files have been modified and reformatted.

veeenu:
Do you have the latest Rust nightly installed? It's unfortunately necessary because some lints have not been stabilized yet.

Thoxy67:
I always use nightly as default target I have tested on Linux this going to also modify format the markdown...

I have made without the `--all` argument this only modify the rust files and seems to have fix the CI lint issue

veeenu:
Tested it, works great, thank you again!

veeenu (approved):

Issue #155, opened 2024-02-14
Errors on Resident Evil 2 and Testing on Other Games


vSylva:
![{7BEBC450-7DF1-4a5c-9C75-36ADAF8AF010}](https://github.com/veeenu/hudhook/assets/58687743/18c61117-6808-43ad-a4ca-a2a6b10cecab)
---

https://github.com/veeenu/hudhook/blob/heck-dcomp/examples/simple_hook.rs

Injecting "simple_hook.dll" after the game's logo appears will result in the aforementioned error. 
Injecting it after the game's main menu appears allows it to run properly. 

OGL game Rune Factory 4, DX11 game Sekiro, DX12 game Resident Evil 3 perform well.

veeenu:
That branch is in a broken state, dx9 and opengl aren't done ye because I can't find a way to share their buffer. I'm working on a Vulkan renderer atm which *should* be able to deal with that iiuc.

vSylva:
Sekiro (DX11) passed the test.
![image](https://github.com/veeenu/hudhook/assets/58687743/164e4e57-c8d2-4023-89c2-feceeeed229e)

Resident Evil 2 (DX12) and Resident Evil 3 (DX12) failed the test; once demo_hook_dx12.dll is injected, it will crash immediately. 
![image](https://github.com/veeenu/hudhook/assets/58687743/04aded38-8e2e-4a93-9c1b-ee83ee9fe4ab)

Rune Factory 4 (OGL3) cannot pass, error: ERROR ThreadId(02) hudhook::hooks::opengl3: src\hooks\opengl3.rs:92: Render error: Error { code: HRESULT(0x80070057), message: "Invalid parameter." }
![image](https://github.com/veeenu/hudhook/assets/58687743/1256491d-e8a2-4e94-947c-5b1435920a04)

veeenu:
Thank you for testing these! I should probably add some observability, I'm not sure why those errors happen.

vSylva:
> Thank you for testing these! I should probably add some observability, I'm not sure why those errors happen.

7bbbe06

(DX12) Resident Evil 2 and Resident Evil 3 appear to work well

![image](https://github.com/veeenu/hudhook/assets/58687743/3bf9e257-d97b-4e7e-9bd3-02281c5d1fd5)
![image](https://github.com/veeenu/hudhook/assets/58687743/9a683488-ea6a-403f-bf37-bc3d5d1070c9)

(DX11) Sekiro as well

[...]

Issue #139, opened 2023-11-18
Crashes while unhooking


OneXDeveloper:
Hi, me again! Been awhile!

I'm still facing crashes frequently while ejecting for DX12 game D2R. I also sometimes get crashes while injecting. I think there might still be a sync issue somewhere. Could you help me debug? What info can I give you that would help?

Here's the last log I get when it crashes during eject. 

```
DEBUG hudhook::hooks::dx12: WndProc called before hook was set
```

If it would be easier to chat over discord, I'd be down to help track these down. They're driving me crazy :P

veeenu:
Hi! Thank you for reporting this!

Unfortunately it's not a great moment for me, between commitments and other personal issues, so I've paused work on hudhook for a bit. I'll still review pull requests though, so if you get to this before I have a chance to, feel free to open one!

The culprit is most likely around here:
https://github.com/veeenu/hudhook/blob/bebcddcb82f462061c1101383001677b4af276ee/src/hooks/dx12.rs#L782-L809

The synchronization is already a hack, and a potential solution is included in the comment, but I (personally) need a better understanding of the DX12 pipeline to debug that. Attaching a debugger to inspect the crash should provide some insight.

Issue #130, opened 2023-08-27
Encountered an issue while using the Example Dx11 Hook with Dx11Game Rune Factory 5


vSylva:
![GIF 2023-8-27 23-19-53](https://github.com/veeenu/hudhook/assets/58687743/77c4db00-eff0-420f-b69c-4430b45acc2a)
===

hudhook rev [3b86352](https://github.com/veeenu/hudhook/commit/3b86352420b37bcc1dc3208334d467837d6230d6)

The game is unable to respond to the mouse and keyboard inputs properly.

No ERROR is displayed on the console.

After pressing a key, the game responds multiple times to the pressed key after a few seconds. Mouse clicks require multiple attempts, and the response is delayed by a few seconds.

From the example dx11_hook
===
```rust
use hudhook::hooks::dx11::ImguiDx11Hooks;
use hudhook::hooks::ImguiRenderLoop;

use tracing::metadata::LevelFilter;
struct Dx11HookExample;

impl Dx11HookExample {
    fn new() -> Self {
        println!("Initializing");
        hudhook::alloc_console().expect("AllocConsole");
        hudhook::enable_console_colors();

        tracing_subscriber::fmt()
            .with_max_level(LevelFilter::ERROR)
            .with_thread_ids(true)
            .with_file(true)
            .with_line_number(true)
            .with_thread_names(true)
            .init();

        Dx11HookExample
    }
}

impl ImguiRenderLoop for Dx11HookExample {
    fn render(&mut self, _ui: &mut imgui::Ui) {
        // ui.window("Hello world").size([300.0, 110.0], Condition::FirstUseEver).build(|| {
        //     ui.text("Hello world!");
        //     ui.text("こんにちは世界！");
        //     ui.text("This...is...imgui-rs!");
        //     ui.separator();
        //     let mouse_pos = ui.io().mouse_pos;
        //     ui.text(format!("Mouse Position: ({:.1},{:.1})", mouse_pos[0], mouse_pos[1]));
        // });
    }
}

hudhook::hudhook!(Dx11HookExample::new().into_hook::<ImguiDx11Hooks>());
```

vSylva:
I have tried multiple methods, but still unable to resolve the issue🤷
I performed hooking using C++ and rendered an ImGui demo window, and did not encounter the aforementioned issues.

vSylva:
> It looks like mouse inputs are being translated incorrectly somewhere.
> 
[...]

Issue #78, opened 2023-02-13
Performance issues in some games in native windowed borderless mode


veeenu:
See [this issue](https://github.com/veeenu/eldenring-practice-tool/issues/37) on the Elden Ring practice tool.

The FPS of the ERPT drops significantly when in windowed borderless mode in at least the issue creator's setup.
We should investigate whether this is consistently reproducible both using in-game windowed borderless and using external tools like Borderless Gaming.

PR #156, opened 2024-02-24
Per-engine compositors

### LINKED ISSUE #151: `DirectComposition` is not implemented on Wine
Using the latest commit, I can't hook into the OpenGL game "Star Stable Online", tested on both Windows and Linux (via Wine).
image

Commits prior to the RenderEngine works.

veeenu:
I come to you, dear community, with a PR that would infuse a merciless sense of dread even in the hearts of the greatest practitioners of Zen meditation alive.

immagine

This PR is close to a full rewrite of `hudhook`'s core. The plan is to carefully review every single file and running extensive tests before smashing the green button, but the quality has in my opinion improved greatly.

# Changes

## Unified renderer

We stay the course with the single `imgui` renderer as of the previous PR #143.

The single renderer allows us to work on `imgui` specific circumstances from a single unified spot. This way, UI rendering issues and new features don't have to be spread over four different renderers, and we will be able in the future to leverage rendering concurrency if we so wish.

This is currently not so; the `imgui` frame is constructed and rendered right in the middle of the `Present` hooks, and blocking those calls. Luckily `hudhook` and `imgui` are pretty slick and can do all that in all four renderers without ever dipping below a solid 100fps on my machine (RTX 3080).

## Per-engine compositors

I took a deep plunge in the dark seas of rendering engines and built four full-screen quad renderers from scratch, one per supported engine, which take the off-screen rendered resource and composite it on top of the appropriate backbuffer.

- The DirectX 12 compositor works seamlessly.
- The DirectX 11 compositor needs to open the off-screen resource as a shared resource.
- The DirectX 9 and OpenGL3 renderers do not have an easy means of sharing memory with DirectX 12, so a very bad fix was necessary: after the DirectX 12 rendering is done, the resource is staged and mapped, its data is pulled on the CPU side and reuploaded to the DirectX 9 / OpenGL3 compositors as their own kind of texture. This is all manners of awful because we move from the GPU to the CPU back to the GPU which has insane latency in theory. In practice, I never dipped below 100fps in all my tests despite of this, and the DirectX 9 and OpenGL 3 games I tested don't seem to bat an eye at all. 
  If you are using one of these renderers, **PLEASE** test this branch to confirm that there is no performance penalty.
  If you are willing to dig for a solution that doesn't leave the GPU, I'd be forever grateful.

## Asynchronous input processing

There is now a `Pipeline` object (its design is still a bit dirty, I want to work on it) that is responsible for managing the lifecycle of the rendering engine, the `imgui` context, and among those operations will also manage the window procedure. The window procedure proper is now dummy and only ships its parameters to a `mpsc::channel`.

This has the advantage that the window events are now processed asynchronously just before a frame gets rendered, which should be fine on our end, but the drawback that an event cannot be blocked immediately should the render loop wish to do that, as it won't be able to decide right away whether to consume the events without forwarding them. This should not be a big deal as these decisions sort of depend on the state before the rendering anyway, rather than on any single specific event, and the previous model is not more fine grained than that.

If you had issues with blocking input, please give this branch a go, it might just inadvertently fix that. Or make things worse. It did not make things worse for my use cases.

## Wine support

I have tested on-and-off the library on Wine with Proton, and it seems like it's working pretty well. There's still work to be done to figure things out, but we're getting there.

## Todo list

- [ ] Refactor the `Pipeline` external API; it is very similar in all renderers and the current solution looks very bad
- [ ] Refactor the window procedures as they are identical across renderers and can be hidden away
- [ ] Review the entire codebase
- [ ] Add documentation and comments where needed

## Affected issues

- Closes #151 
- #155 needs review
- #139 needs review
- #137 needs review
- #130 needs review
- #107 needs review
- #103 needs review
- #96 needs review
- #79 needs review
- #78 needs review

# Conclusion

All in all, I am very happy about the results of this PR and have been working tirelessly towards it. There are some outstanding issues, but all manageable. I will keep making inconsequential changes while I wait for your feedback, but my plan is to wrap this up sooner rather than later and finally release `0.6.0`.

cc @Jakobzs @vars1ty @vSylva @camas


vars1ty:
Tested in Star Stable Online with the example OGL3 DLL, built with MSVC. All tested via Wine.

Average FPS without, and with UI displayed:
- Before: ~202 FPS
- After: ~60 FPS

So does tank a bit.

src/pipeline.rs
```diff
@@ -0,0 +1,174 @@
...
+        ))
+    }
+
+    pub fn render(&mut self) -> Result<ID3D12Resource> {
+        // TODO find a better alternative than allocating each frame
+        let message_queue = self.rx.try_iter().collect::<Vec<_>>();
+
+        message_queue.into_iter().for_each(|PipelineMessage(hwnd, umsg, wparam, lparam)| {
+            imgui_wnd_proc_impl(hwnd, umsg, wparam, lparam, self);
+        });
```
camas, line 96: Pass directly to for_each?

```suggestion
        self.rx.try_iter().for_each(|PipelineMessage(hwnd, umsg, wparam, lparam)| {
            imgui_wnd_proc_impl(hwnd, umsg, wparam, lparam, self);
        });
```

veeenu: Unfortunately not: `self.rx.try_iter()` already mutably borrows `self`, and `imgui_wnd_proc_impl` would also have to mutably borrow `self` for the same lifetime which is not possible.

Similarly, allocating a vector in `Pipeline` and draining/extending it there with the contents of `try_iter` is also not going to work for the same reason, barring using `RefCell` which has overhead anyway -- maybe less so than allocating a vector outright, but at the cost of making the code look a lot uglier.

veeenu:
This PR feels more like the development of a redemption arc rather than good, conscientious software engineering, but here goes.

The performance penalty incurred by @vars1ty is not acceptable, so I don't see this copy-buffers solution as viable.

Reviewing my own code, I realized that the full screen quad renderers which I wrote from scratch are literally 30-40 lines away each from being complete, fully-fledged `dear imgui` renderers. The difference is literally just uploading data to vertex/index buffers and adding render commands according to an enum match in a loop.

In the process, I have acquired enough knowledge about the various rendering engines and GPUs that I now feel confident maintaining all of these things by myself.

In fact, I went back and checked the renderers' code from 0.5.0 and I _shuddered_. What the heck was that garbage, @veeenu !

We will thus go back to the previous model, skipping the current PR: one rendering engine per hook type.

## The way forward: considerations

- The windowing event behavior was much different across hooks and required a lot more code in the legacy way compared to the current solution. All of that behavior salad is now already accounted for, and improved upon, by the `pipeline` module.

- Reverting to per-renderer renderers will come at the cost of not being able to do concurrent rendering in `opengl3` and `dx9`, which just don't have the capability (that I know of) of easily working asynchronously. For the time being, neither the `dx11` and `dx12` renderers will have that capability, though that is true of this current PR as well; but it is still potentially viable to do so at some point in the future, should that need arise.

- I'm hoping that all in all the new, from-scratch renderers are going to be a big, net decrease in code, and an overall simplification and streamlining of behavior, to the point where there might be no need to touch them at all for maintenance and having four of them won't be burdensome at all.

- The `imgui` context will be managed externally by `Pipeline` (like as of this PR) and only mutably passed to the renderers for a limited lifetime, which is a _great_ improvement in ergonomics and separation of concerns compared to the legacy approach where the renderer created and owned the context altogether.

## Conclusions

More updates to come soon. I have a plan well laid out in my mind to make good out of this absolute mess that I have willingly put myself in the midst of. February is coming to an end, which means that if everything goes according to plan, I will have a solid solution by the time it will be both feasible and advisable for me to finally go out and touch grass. 🌱

src/pipeline.rs
veeenu: I have concocted a solution for this:

- Store an allocated `Vec` in a `OnceCell`.
- Pull it out of the `OnceCell` before this section and use it as a backing storage for `rx.try_iter()`.
- Drain it immediately (preserving the capacity) and pass the contents to the wnd proc implementation.
- Store the drained `Vec` back in the `OnceCell`.

This is a bit ugly, but works around the lifetime issues with little to no performance overhead -- `OnceCell::take` and `OnceCell::set` are simple moves, memory is only ever reallocated when needed, a generous buffer could be pre-allocated, and even a large vector won't be that much of a memory waste.

veeenu:
Superseded by #158.

PR #176, opened 2024-03-14
Replace texture image data

### LINKED ISSUE #147: Replace texture image data
Currently, a texture can only be loaded once via `RenderEngine::load_image` and its contents cannot be changed.

A public `RenderEngine::replace_image` method should be created to make this happen.

misdake:
Fixes #147.

1. add TextureLoader trait with load_texture and replace_texture
2. implement TextureLoader for all 4 engines
    * dx9 and dx12: load_texture = heap.create_texture + heap.upload_texture, replace_texture = heap.upload_texture
    * dx11 and gl3: load_texture = heap.create_texture (with init), replace_texture = heap.update_texture

"Allow edits by maintainers" is on. If you have any suggestions for improvement, just push changes to this branch, specially about naming and code repetition.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -143,9 +143,21 @@ static mut MODULE: OnceCell<HINSTANCE> = OnceCell::new();
 static mut HUDHOOK: OnceCell<Hudhook> = OnceCell::new();
 static CONSOLE_ALLOCATED: AtomicBool = AtomicBool::new(false);
 
-/// A load texture callback. Invoke it in your
-/// [`crate::ImguiRenderLoop::initialize`] method for setting up textures.
-pub type TextureLoader<'a> = &'a mut dyn FnMut(&'a [u8], u32, u32) -> Result<TextureId, Error>;
+/// Texture Loader for ImguiRenderLoop callbacks to load and replace textures
+pub trait TextureLoader {
+    /// Load texture and return TextureId to use. Invoke it in your
+    /// [`crate::ImguiRenderLoop::initialize`] method for setting up textures.
+    fn load_texture(&mut self, data: &[u8], width: u32, height: u32) -> Result<TextureId, Error>;
+    /// Upload an image to an existing texture, replacing its content. Invoke it in your
+    /// [`crate::ImguiRenderLoop::before_render`] method for updating textures.
+    fn replace_texture(
+        &mut self,
+        texture_id: TextureId,
+        data: &[u8],
+        width: u32,
+        height: u32,
+    ) -> Result<(), Error>;
+}
 
 /// Allocate a Windows console.
 pub fn alloc_console() -> Result<(), Error> {
@@ -219,14 +231,24 @@ pub fn eject() {
 pub trait ImguiRenderLoop {
     /// Called once at the first occurrence of the hook. Implement this to
     /// initialize your data.
-    fn initialize<'a>(&'a mut self, _ctx: &mut Context, _loader: TextureLoader<'a>) {}
+    fn initialize<'a>(
+        &'a mut self,
+        _ctx: &mut Context,
+        _texture_loader: &'a mut dyn TextureLoader,
+    ) {
+    }
 
     /// Called every frame. Use the provided `ui` object to build your UI.
     fn render(&mut self, ui: &mut Ui);
 
     /// Called before rendering each frame. Use the provided `ctx` object to
     /// modify imgui settings before rendering the UI.
-    fn before_render(&mut self, _ctx: &mut Context) {}
+    fn before_render<'a>(
+        &'a mut self,
+        _ctx: &mut Context,
+        _texture_loader: &'a mut dyn TextureLoader,
+    ) {
+    }
 
     /// Called during the window procedure.
     fn on_wnd_proc(&self, _hwnd: HWND, _umsg: u32, _wparam: WPARAM, _lparam: LPARAM) {}
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -148,6 +148,7 @@ pub trait RenderContext {
     /// Load texture and return TextureId to use. Invoke it in your
     /// [`crate::ImguiRenderLoop::initialize`] method for setting up textures.
     fn load_texture(&mut self, data: &[u8], width: u32, height: u32) -> Result<TextureId, Error>;
+
     /// Upload an image to an existing texture, replacing its content. Invoke it
     /// in your [`crate::ImguiRenderLoop::before_render`] method for
     /// updating textures.
@@ -232,6 +233,8 @@ pub fn eject() {
 pub trait ImguiRenderLoop {
     /// Called once at the first occurrence of the hook. Implement this to
     /// initialize your data.
+    /// `ctx` is the imgui context, and `render_context` is meant to access
+    /// hudhook renderers' extensions such as texture management.
     fn initialize<'a>(
         &'a mut self,
         _ctx: &mut Context,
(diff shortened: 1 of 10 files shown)


misdake:
I'd say tests are now scary as hell.
image

src/renderer/pipeline.rs
```diff
@@ -110,7 +108,7 @@ impl<T: RenderEngine> Pipeline<T> {
         io.nav_active = true;
         io.nav_visible = true;
 
-        self.render_loop.before_render(&mut self.ctx);
+        self.render_loop.before_render(&mut self.ctx, &mut self.engine);
```
veeenu, line 111: I like this trait based approach.

I originally went for the closure type because I was running into borrow checker issues: over here, `self`, `self.render_loop` and `self.engine` are all mutably borrowed at the same time and the fact that it is compiling at all is kinda throwing me for a loop, but I'm gonna trust Rust on this.

I wonder if `TextureLoader` is the right name or we might just abstract it a little bit further and make it a more generic public `RenderEngine` trait (separated from `crate::renderer::RenderEngine` which should stay private, but also I'd like to find a name which makes it clear that they are distinct and one is meant to be private) so we can add more methods in the future in a non-breaking fashion.

This would be beneficial for covering other future use cases like exposing renderers' internals (to e.g. fetch projection matrices/textures from the game's renderers).

veeenu (commented):
That's a scary swole doge if I've ever seen one. 😅

Style looks good to me in general, please allow me a bit of time so I can review it properly but it's going in the right direction.

This will probably warrant a breaking semver minor bump to 0.7 since the public API is getting changed.

Thank you!

src/renderer/pipeline.rs
misdake: “RenderResource”, or "RenderContext/Ctx"？ImguiRenderLoop::render can also have this in param list without issues.

misdake: Yeah, I'm quite against adding it to `render`. I like the name `RenderContext`, it's just kinda strange to have `RenderContext` in `init` + `before_render` but not in `render`.

src/lib.rs
```diff
@@ -219,14 +232,24 @@ pub fn eject() {
 pub trait ImguiRenderLoop {
     /// Called once at the first occurrence of the hook. Implement this to
     /// initialize your data.
-    fn initialize<'a>(&'a mut self, _ctx: &mut Context, _loader: TextureLoader<'a>) {}
-
-    /// Called every frame. Use the provided `ui` object to build your UI.
-    fn render(&mut self, ui: &mut Ui);
+    fn initialize<'a>(
```
veeenu, line 235: Could you please adapt the comment here stating that `ctx` is the imgui context and `render_context` is meant to access hudhook renderers' extensions such as texture management?

src/lib.rs
```diff
@@ -219,14 +232,24 @@ pub fn eject() {
...
+        &'a mut self,
+        _ctx: &mut Context,
+        _render_context: &'a mut dyn RenderContext,
+    ) {
+    }
 
     /// Called before rendering each frame. Use the provided `ctx` object to
     /// modify imgui settings before rendering the UI.
-    fn before_render(&mut self, _ctx: &mut Context) {}
+    fn before_render<'a>(
```
veeenu, line 244: Same as above

src/lib.rs
```diff
@@ -143,9 +143,22 @@ static mut MODULE: OnceCell<HINSTANCE> = OnceCell::new();
...
 
-/// A load texture callback. Invoke it in your
-/// [`crate::ImguiRenderLoop::initialize`] method for setting up textures.
-pub type TextureLoader<'a> = &'a mut dyn FnMut(&'a [u8], u32, u32) -> Result<TextureId, Error>;
+/// Texture Loader for ImguiRenderLoop callbacks to load and replace textures
+pub trait RenderContext {
+    /// Load texture and return TextureId to use. Invoke it in your
+    /// [`crate::ImguiRenderLoop::initialize`] method for setting up textures.
+    fn load_texture(&mut self, data: &[u8], width: u32, height: u32) -> Result<TextureId, Error>;
+    /// Upload an image to an existing texture, replacing its content. Invoke it
```
veeenu, line 151: Could you please add a newline here above the comment?

src/renderer/backend/dx11.rs
```diff
@@ -612,12 +625,40 @@ impl TextureHeap {
...
+    unsafe fn update_texture(
+        &mut self,
+        texture_id: TextureId,
+        data: &[u8],
+        width: u32,
+        height: u32,
+    ) -> Result<()> {
+        let texture = &mut self.textures[texture_id.id()];
+        if texture.width != width || texture.height != height {
+            return Err(Error::new(
```
veeenu, line 642: I'd rather not construct the errors this way: they call out to a function which is not always available and leads to crashes on Wine, check out #167. 

This is not ideal since it relies on logging, but I'd use the following idiom instead:

```rs
error!(format!(/* ... */));
return Err(Error::from_hresult(HRESULT(-1));
```

In the future I'd like to go with a custom error format but for now this'll do since it's used only in a few spots.

src/renderer/backend/dx12.rs
```diff
@@ -750,6 +762,48 @@ impl TextureHeap {
...
+    unsafe fn upload_texture(
+        &mut self,
+        texture_id: TextureId,
+        data: &[u8],
+        width: u32,
+        height: u32,
+    ) -> Result<()> {
+        let texture = &self.textures[texture_id.id()];
+        if texture.width != width || texture.height != height {
+            return Err(Error::new(
```
veeenu, line 798: Same observation as above, just marking the spot with a  comment for convenience.

src/renderer/backend/dx9.rs
```diff
@@ -384,8 +396,32 @@ impl TextureHeap {
...
+    unsafe fn upload_texture(
+        &mut self,
+        texture_id: TextureId,
+        data: &[u8],
+        width: u32,
+        height: u32,
+    ) -> Result<()> {
+        let texture = &self.textures[texture_id.id()];
+        if texture.width != width || texture.height != height {
+            return Err(Error::new(
```
veeenu, line 414: Same observation as above, just marking the spot with a  comment for convenience.

src/renderer/backend/opengl3.rs
```diff
@@ -375,9 +392,52 @@ impl TextureHeap {
...
+        &mut self,
+        gl: &gl::Gl,
+        texture: TextureId,
+        data: &[u8],
+        width: u32,
+        height: u32,
+    ) -> Result<()> {
+        let texture_info = self.get(texture);
+        if texture_info.width != width || texture_info.height != height {
+            return Result::Err(Error::new(
```
veeenu, line 411: Same observation as above, just marking the spot with a  comment for convenience.

(is the `Result::` prefix necessary here?)

veeenu (changes requested):
Thank you, this is great! Works fine on Wine too.

I have just a couple observations/nits for style, and one important request for compat.

Note: After this gets merged in, we'll have to bump to `0.7.0` due to breaking changes and update `hudbook` accordingly.

veeenu (approved):
LGTM! Thank you so much :raised_hands:

Godnoken:
I've been trying to implement similar functionality from inside the renderloop (I am passing the swapchain to it), using hudhook 0.4(?), and only managed to do so for OpenGL so far. DX9 crashes no matter what, DX11 displays random white symbols (what?) and DX12 is not tested. I'm not entirely sure if it is even possible without modifying the hudhook code.

From what I can tell, the current functionality on hudhook 0.8 only allows the upload of textures from the initializing stage. Is this correct?

veeenu:
`render` takes a frame object that immutably borrows from `Context`, which then makes it impossible to mutably borrow from it in order to upload textures. `before_render` is called back just before that happens and gives access to a mutable reference to `Context`. Textures can be uploaded and modified any time in `before_render` -- it gets called for every frame, and it requires some state tracking, but of course it's not advisable to actually swap textures at every frame. :sweat_smile:

PR #179, opened 2024-03-29
Let minhook be a Git submodule

vSylva:
Maybe this way would be better?


vSylva:
The suggested method is not suitable. If the person's computer does not have git, then it is not possible to compile hudhook.

veeenu:
The existence of `git` is not a big deal -- it's a reasonable dependency not to mention some parts of `cargo` itself rely on it. But I still think vendoring the code is a better idea: a submodule means relying on an external data source that's outside of our control, and that disadvantage is not offset by any equal or superior advantage to us, especially considering how small and self contained `minhook` is.

PR #183, opened 2024-04-16
add MessageFilter to selectively block window message

### LINKED ISSUE #182: need ways to filter certain windows message, without blocking everything
My payload uses `should_block_message()` to prevent propagating keyboard input to game when my overlay should handle one.
For my target game there's no problem about that (It seems the target only uses wndproc to get keyboard event).

Recently I tried to use [imgui-rs hello_world example window](https://github.com/imgui-rs/imgui-rs/blob/main/imgui-examples/examples/hello_world.rs) as a test rig for my overlay UI, and quickly got into trouble.

Turns out the example program uses `WM_PAINT` message as render logic entry.
For that situation, if `should_block_message()` got set, next frame will be unreachable.
(Filter and unsafely send message back on `on_wnd_proc()` approach is impossible due to that.)

Since current "block them all" thing seems very limiting and dangerous in some cases, I think we need other approach.

ruby3141:
This PR introduces BREANKING changes described below;
- add MessageFilter to selectively block window message
  Blocking every window message blocks intentional user actions like minimizing, closing(Alt-F4 doesn't work either), resizing, etc.
  This commit fixes that kind of problem by providing bitflag interface to block certain predefined range of window messages.
  To block every messages like before, you can return `MessageFilter::all()` on `set_message_filter()`.

This PR fixes #182.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -261,9 +261,9 @@ pub trait ImguiRenderLoop {
     /// Called during the window procedure.
     fn on_wnd_proc(&self, _hwnd: HWND, _umsg: u32, _wparam: WPARAM, _lparam: LPARAM) {}
 
-    /// Sets the types of window message that
+    /// Returns the types of window message that
     /// you do not want to propagate to the main window
-    fn set_message_filter(&self, _io: &Io) -> MessageFilter {
+    fn message_filter(&self, _io: &Io) -> MessageFilter {
         MessageFilter::empty()
     }
 }
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -136,6 +136,8 @@ pub mod inject;
 pub mod mh;
 pub(crate) mod renderer;
 
+pub use renderer::msg_filter::MessageFilter;
+
 pub mod util;
 
 // Global state objects.
@@ -259,10 +261,10 @@ pub trait ImguiRenderLoop {
     /// Called during the window procedure.
     fn on_wnd_proc(&self, _hwnd: HWND, _umsg: u32, _wparam: WPARAM, _lparam: LPARAM) {}
 
-    /// If this method returns `true`, the WndProc function will not call the
-    /// procedure of the parent window.
-    fn should_block_messages(&self, _io: &Io) -> bool {
-        false
+    /// Sets the types of window message that
+    /// you do not want to propagate to the main window
+    fn set_message_filter(&self, _io: &Io) -> MessageFilter {
+        MessageFilter::empty()
     }
 }
 
--- a/src/renderer/msg_filter.rs
+++ b/src/renderer/msg_filter.rs
@@ -44,7 +44,7 @@ bitflags! {
 
 impl MessageFilter {
     /// Check whether the message ID is blocked by this filter
-    pub(crate) fn is_blocking(&self, message_id: u32, _wparam: usize, _lparam: isize) -> bool {
+    pub(crate) fn is_blocking(&self, message_id: u32) -> bool {
         if match message_id {
             0x0000..=0x03FF => self.contains(Self::RangeSystemDefined),
             WM_USER..=0x7FFF => self.contains(Self::RangePrivateReserved),
--- a/src/renderer/msg_filter.rs
+++ b/src/renderer/msg_filter.rs
@@ -1,0 +1,74 @@
+//! This module contains logic for filtering windows messages.
+
+use bitflags::bitflags;
+use windows::Win32::UI::WindowsAndMessaging::*;
+
+bitflags! {
+    /// Flags for filtering window messages.
+    #[repr(transparent)]
+    pub struct MessageFilter: u32 {
+        /// Blocks keyboard input event messages.
+        const InputKeyboard = 1u32 << 0;
+        /// Blocks mouse input event message.
+        const InputMouse = 1u32 << 1;
+        /// Blocks raw input event messages.
+        const InputRaw = 1u32 << 2;
+
+        /// Blocks window gain/lose focus event messages.
+        const WindowFocus = 1u32 << 8;
+        /// Blocks window control event messages
+        /// like move, resize, minimize, etc.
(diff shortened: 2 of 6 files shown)


veeenu:
Thank you for your contribution! 🙏 
Please allow me a bit of time to thoroughly review this, but I like what I've seen so far.

src/renderer/msg_filter.rs
```diff
@@ -0,0 +1,74 @@
...
+        /// Blocks keyboard, mouse, raw input messages.
+        const InputAll = Self::InputKeyboard.bits() | Self::InputMouse.bits() | Self::InputRaw.bits();
+        /// Blocks window focus, control, close messages.
+        const WindowAll = Self::WindowFocus.bits() | Self::WindowControl.bits() | Self::WindowClose.bits();
+    }
+}
+
+impl MessageFilter {
+    /// Check whether the message ID is blocked by this filter
+    pub fn is_blocking(&self, message_id: u32, _wparam: usize, _lparam: isize) -> bool {
```
veeenu, line 47: `_wparam` and `_lparam` aren't used here. Considering that this is a public facing API, this will be confusing to users. Could you please remove them?

src/renderer/msg_filter.rs
```diff
@@ -0,0 +1,74 @@
+//! This module contains logic for filtering windows messages.
+
+use bitflags::bitflags;
+use windows::Win32::UI::WindowsAndMessaging::*;
+
+bitflags! {
+    /// Flags for filtering window messages.
```
veeenu, line 7: Could you please add a long-form explanation to this, with examples, so users can get a better idea at a glance of what this data structure is for and how it should be used?

src/lib.rs
```diff
@@ -259,10 +261,10 @@ pub trait ImguiRenderLoop {
     /// Called during the window procedure.
     fn on_wnd_proc(&self, _hwnd: HWND, _umsg: u32, _wparam: WPARAM, _lparam: LPARAM) {}
 
-    /// If this method returns `true`, the WndProc function will not call the
-    /// procedure of the parent window.
-    fn should_block_messages(&self, _io: &Io) -> bool {
-        false
+    /// Sets the types of window message that
+    /// you do not want to propagate to the main window
+    fn set_message_filter(&self, _io: &Io) -> MessageFilter {
```
veeenu, line 266: Could we change the name here? `set_` prefix usually implies some sort of internal mutation, whereas this method returns a value. Just `message_filter` could work, but I'm willing to consider other ideas.

What's important is that the user sees the method name and immediately thinks "ok, here I need to implement the logic that will tell the system which messages to ignore according to the current state".

src/renderer/msg_filter.rs
```diff
@@ -0,0 +1,74 @@
...
+
+        /// Blocks window gain/lose focus event messages.
+        const WindowFocus = 1u32 << 8;
+        /// Blocks window control event messages
+        /// like move, resize, minimize, etc.
+        const WindowControl = 1u32 << 9;
+        /// Blocks window close messages.
+        const WindowClose = 1u32 << 10;
+
+        /// Blocks messages ID from 0 to WM_USER - 1
```
veeenu, line 25: Could you please add backticks to the constants here e.g. `WM_USER`? Just for the sake of documentation consistency.

veeenu (changes requested):
Thank you for your contribution. I like this!

I'd like to have a few changes but otherwise this looks good. Also, could you please implement the new method in the [test hook](https://github.com/veeenu/hudhook/blob/main/tests/hook.rs)? This would be helpful both to users looking for an example and to us to check that the thing works correctly. Could e.g. ignore events according to time elapsed from `first_time` (2s ignore keyboard, 4s ignore mouse etc).

src/renderer/msg_filter.rs
ruby3141: It was "Here's your starting point if you want to fork and implement custom complex filter" kind of parameters,
but I'm not confident about its necessity either.
If you don't agree with its reason of existance, I'll remove it. Please let me know.

veeenu: I don't think ease of forking should be a goal. It would be about a separate project after all, in a sense, and it's unreasonable to optimize for that and document that. Plus, it might be confusing to the readers of the code. Yeah, I'd say let's remove them. Thank you!

ruby3141:
In addition to your change requests, I made `MessageFilter::is_blocking()` `pub(crate)`,
because it would not be used in user code.

Test for only `MessageFilter::WindowControl` is added to the test hook, because harness itself doesn't seems to handle any input.

ruby3141:
[out.webm](https://github.com/veeenu/hudhook/assets/3256382/704b79e4-dbde-4d4d-82dd-b08c649d9d98)

veeenu (approved):
This looks great, thank you again! 🙌 

(p.s. what font is that in your editor? I like it!)

ruby3141:
The font is [NeoDunggeunmo Code](https://github.com/neodgm/neodgm/blob/main/README.en.md).
It's one of a few programming fonts that natively supports korean glyphs, and because of its bitmap appearance and square sizing it matches very well with unifont fallback.

Godnoken:
Oi oi!

I've just recently started working on my application again.

I'm quite hesitant to spend time bumping up since I ended up changing a lot of code in my own local hudhook repo and I haven't had the chance to try newer versions of hudhook yet, but I'm wondering if this PR made it possible to completely block all input to games?

P.S.
I've been following the development with one eye for over 1.5 years, happy to see that it has been going strong! :)

veeenu:
Hey, welcome back @Godnoken! :raised_hands: 

This specifically did not solve my own issue with Souls games but the likely reason is that the input isn't sourced via windows events at all, nor `GetKeyState`, and it's instead directinput or something like that which I haven't looked into yet. Games that do that should for now add their own hook implementations, I'm not opposed to upstreaming that kind of functionality for common cases though.

PR #184, opened 2024-04-28
Bugfix panic in D3D11RenderEngine->restore

FrankvdStam:
When self.ps_instances is none, self.ps_instances_count is 0, then slice::from_raw_parts causes a panic here.
Same for the vs version. Initially I tried this fix with is_some() but it still caused a crash for me.

I don't know if this has any side effects, for me it renders fine after this fix.

### DIFF
--- a/src/renderer/backend/dx11.rs
+++ b/src/renderer/backend/dx11.rs
@@ -821,13 +821,10 @@ impl StateBackup {
             device_context.PSSetSamplers(0, Some(&self.sampler));
         }
 
-        if self.ps_instances_count > 0
-        {
-            device_context.PSSetShader(
-                self.pixel_shader.as_ref(),
-                Some(slice::from_raw_parts(self.ps_instances, self.ps_instances_count as usize)),
-            );
-        }
+        device_context.PSSetShader(
+            self.pixel_shader.as_ref(),
+            if self.ps_instances_count > 0 { Some(slice::from_raw_parts(self.ps_instances, self.ps_instances_count as usize)) } else { None },
+        );
 
         if self.vs_instances_count > 0
         {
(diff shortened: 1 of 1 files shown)


src/renderer/backend/dx11.rs
```diff
@@ -821,14 +821,21 @@ impl StateBackup {
...
-            self.vertex_shader.as_ref(),
-            Some(slice::from_raw_parts(self.vs_instances, self.vs_instances_count as usize)),
-        );
+        if self.ps_instances_count > 0
+        {
+            device_context.PSSetShader(
+                self.pixel_shader.as_ref(),
+                Some(slice::from_raw_parts(self.ps_instances, self.ps_instances_count as usize)),
+            );
+        }
```
veeenu, line 830: Since we want to restore the previous state, it would probably be a better idea to call it regardless with null arguments, otherwise the game would see our shaders:

```suggestion
        device_context.PSSetShader(
            self.pixel_shader.as_ref(),
            if self.ps_instances_count > 0 { Some(slice::from_raw_parts(self.ps_instances, self.ps_instances_count as usize)) } else { None },
        );
```

(same for vertex shaders below)

veeenu (changes requested):

FrankvdStam:
Committed your suggestion and did the same bellow, plus fixed the linting error, tested and it works on remastered.

veeenu:
Oh, right, CI fails because of the new `transmute` lint. I'll fix that first and then merge this one, shouldn't need a rebase hopefully.

FrankvdStam:
This change is so small, I don't think there will be any issues merging

veeenu (approved):
LGTM, thank you :heart:

Issue #186, opened 2024-05-16
hudhook pipeline doesn't report `delta_time` to imgui runtime


ruby3141:
According to imgui's custom backend implementation exhibit and first-party backend implementations,
`io.DeltaTime` is "usually" meant to be updated by backend code.

https://github.com/ocornut/imgui/blob/93daf232233dd0a68c6b0404d53b18e0d089b0d1/imgui.cpp#L334-L336
https://github.com/ocornut/imgui/blob/93daf232233dd0a68c6b0404d53b18e0d089b0d1/backends/imgui_impl_win32.cpp#L398
https://github.com/ocornut/imgui/blob/93daf232233dd0a68c6b0404d53b18e0d089b0d1/backends/imgui_impl_sdl3.cpp#L691

I think it's good to follow the convention for hudhook's custom imgui backend.

veeenu:
That's a very good idea!

PR #187, opened 2024-05-16
make pipeline reports `delta_time` to imgui

ruby3141:
According to imgui's custom backend implementation exhibit
and first-party backend implementations,
`io.DeltaTime` is "usually" meant to be updated by backend code.

(you can check platform backend impls of ocornut/imgui for example)

This PR solves issue #186.

### DIFF
--- a/src/renderer/pipeline.rs
+++ b/src/renderer/pipeline.rs
@@ -3,6 +3,7 @@ use std::mem;
 use std::sync::atomic::{AtomicU32, Ordering};
 use std::sync::mpsc::{self, Receiver, Sender};
 use std::sync::Arc;
+use std::time::{Duration, Instant};
 
 use imgui::Context;
 use once_cell::sync::{Lazy, OnceCell};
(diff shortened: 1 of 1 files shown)


veeenu (approved):
I'm not super sure what happens when the delta time ends up being `Duration::ZERO`, but it seems to be working. LGTM. Thank you for your contribution!

ruby3141:
imgui prohibits `IO.DeltaTime` of 0 when it's not in its first frame, \
but `Duration::ZERO` will be boundary checked by `imgui::Io::update_delta_time()` into `f32::MIN_POSITIVE` seconds.
https://github.com/ocornut/imgui/blob/a1566c5e1ba22755c359e3079f5f25ab53f1eafb/imgui.cpp#L9784
https://github.com/imgui-rs/imgui-rs/blob/cb8beeb74cceec787275ec3e121c0068ba08fbf7/imgui/src/io.rs#L435-L437

I've considered `Duration::from_nanos(1)` instead of `Duration::ZERO`, but turns out it was too large.

`f32::MIN_POSITIVE` seconds is small enough not to actually increment global timer(which is f64 btw) while being positive.
it is about 0.000000000000000000000000000000000000011754944 seconds, or 0.00000001 quectoseconds,
and when added into f64 global timer, which is usually larger than 1, it just rounded down into zero.
(In f64, the very next value after 1 is 1.0000000000000002. It need about 30 more zeros.)

veeenu:
That's great, thank you!

Issue #191, opened 2024-06-09
Unexpected DirectX9 behavior


Alex141642:
Hello,
First of all, thanks for this library! 
I am opening this issue because the DirectX9 hook does not seem to be behaving correctly when the PC stops rendering the window.
When rendering on the window is not possible, such as:
- Pressing ctrl+alt+del
- Window changing resolution
- Screen going into sleep mode

This causes an error on hudhook: 
```
ERROR hudhook::hooks::dx9 - Render error: Error { code: HRESULT(0x8876086C), message: "" }
```

This error appears to be caused here: `src/renderer/backend/dx9`: `self.render_draw_data(draw_data)?;`

Once the error has occurred, the main window no longer updates, leading me to believe it is a panic.

When testing with openGL, it works as expected.

Here is the code for the menu:
```rust
pub struct Menu {}
impl Menu {
    pub fn new() -> Self {
        Self {}
    }
}
impl ImguiRenderLoop for Menu {
    fn message_filter(&self, io: &hudhook::imgui::Io) -> hudhook::MessageFilter {
        let mut filter = MessageFilter::empty();
        let pos = io.mouse_pos;
        if pos[0] > 0. && pos[0] < 420. && pos[1] > 200. && pos[1] < 600. {
            filter = MessageFilter::InputMouse;
        }
        return filter;
    }
    fn render(&mut self, ui: &mut Ui) {
        ui.window("Demo")
            .movable(false)
            .position([0., 200.], Condition::FirstUseEver)
            .size([420., 400.], Condition::FirstUseEver)
            .resizable(false)
            .build(|| {
                ui.text("Hello world")
            });
    }
}
```
Do you have any idea what might be causing this?

Alex141642:
I made a draft PR wich seems to prevent the menu to disapear. Hower the code is ugly and not fully fonctional.

Alex141642:
Hi @veeenu 
After a lot of try, i can't find a way to make it work properly.
I tried with differents DirectX  libs i found (c/c++/rust), and they all have the same behavour.

veeenu:
Might that behavior depend on the implementation of the host application? Can you reproduce the behavior with a different dx9 application? If we confirm that the behavior is application dependent, I wouldn't upstream the change, as the aim of the library is staying generic. But if we discover the opposite, I'd go ahead and adjust and merge the PR.

Alex141642:
I'll take a look

lucastavares06:
any updates? im having this issue as well

PR #192, opened 2024-06-10
prevent crash when d3d9 render in background

Alex141642:
I opened the issue #191  because pressing keys **ctrl+alt+supr** or **alt+tab** made my D3D9 imgui menu to disapear.

After looking at this topic: [256699-alt-tab-support-directx.html](https://www.unknowncheats.me/forum/c-and-c-/256699-alt-tab-support-directx.html), i noticed that the  WM_ACTIVATEAPP was not handled. I made a little POC that call the Reset method on the device. My menu is still present after that, even if there is rendering errors in the console.

This PR should not be merged directly

### DIFF
--- a/src/renderer/backend/dx9.rs
+++ b/src/renderer/backend/dx9.rs
@@ -79,6 +79,7 @@ impl RenderContext for D3D9RenderEngine {
 
     fn replace_texture(
         &mut self,
+
         texture_id: TextureId,
         data: &[u8],
         width: u32,
@@ -90,7 +91,17 @@ impl RenderContext for D3D9RenderEngine {
 
 impl RenderEngine for D3D9RenderEngine {
     type RenderTarget = IDirect3DSurface9;
-
+    fn reset(&mut self) {
+        tracing::error!("Calling reset stuff");
+        let mut params = D3DPRESENT_PARAMETERS { ..Default::default() };
+        unsafe {
+            let result = self.device.Reset(&mut params as *mut D3DPRESENT_PARAMETERS);
+            match result {
+                Ok(_) => tracing::info!("All is good"),
+                Err(err) => tracing::error!("Error from reset: {}", err),
+            }
+        }
+    }
     fn render(
         &mut self,
         draw_data: &imgui::DrawData,
--- a/src/renderer/input.rs
+++ b/src/renderer/input.rs
@@ -339,6 +339,10 @@ pub fn imgui_wnd_proc_impl<T: RenderEngine>(
         WM_SIZE => {
             pipeline.resize(loword(lparam as u32) as u32, hiword(lparam as u32) as u32);
         },
+        WM_ACTIVATEAPP => {
+            tracing::error!("WM_ACTIVATEAPP trigered, calling reset from engine");
+            pipeline.engine.reset();
+        },
         _ => {},
     };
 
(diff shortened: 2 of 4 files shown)


src/renderer/backend/dx9.rs
```diff
@@ -90,7 +91,17 @@ impl RenderContext for D3D9RenderEngine {
 
 impl RenderEngine for D3D9RenderEngine {
     type RenderTarget = IDirect3DSurface9;
-
+    fn reset(&mut self) {
+        tracing::error!("Calling reset stuff");
```
veeenu, line 95: I'd rather use `tracing::trace!` for this, as this is not an error.

To keep in line with the general crate style, could you please add a `use tracing::{error, ...}` statement at the top? Same for the instances below.

src/renderer/backend/dx9.rs
```diff
@@ -90,7 +91,17 @@ impl RenderContext for D3D9RenderEngine {
...
+        let mut params = D3DPRESENT_PARAMETERS { ..Default::default() };
+        unsafe {
+            let result = self.device.Reset(&mut params as *mut D3DPRESENT_PARAMETERS);
+            match result {
+                Ok(_) => tracing::info!("All is good"),
+                Err(err) => tracing::error!("Error from reset: {}", err),
+            }
```
veeenu, line 102: I'd rather not report the success case at all, and only manage the error case instead. I'd also use the single `e` character name out of style consistency with the codebase.

```suggestion
            if let Err(e) = self.device.Reset(&mut params as *mut D3DPRESENT_PARAMETERS) {
                Err(err) => tracing::error!("Error from reset: {}", err),
            }
```

src/renderer/backend/dx9.rs
```diff
@@ -79,6 +79,7 @@ impl RenderContext for D3D9RenderEngine {
 
     fn replace_texture(
         &mut self,
+
```
veeenu, line 82:
```suggestion
```

src/renderer/input.rs
```diff
@@ -339,6 +339,10 @@ pub fn imgui_wnd_proc_impl<T: RenderEngine>(
         WM_SIZE => {
             pipeline.resize(loword(lparam as u32) as u32, hiword(lparam as u32) as u32);
         },
+        WM_ACTIVATEAPP => {
+            tracing::error!("WM_ACTIVATEAPP trigered, calling reset from engine");
+            pipeline.engine.reset();
```
veeenu, line 344: Similar to above, I wouldn't report this event at all as it is explicitly necessary, and also I'd change this to be a method of `pipeline` instead, so we can avoid making the `engine` field `pub`:
```suggestion
            pipeline.reset();
```

src/renderer/pipeline.rs
```diff
@@ -41,7 +41,7 @@ pub(crate) struct PipelineSharedState {
 pub(crate) struct Pipeline<T: RenderEngine> {
     hwnd: HWND,
     ctx: Context,
-    engine: T,
+    pub engine: T,
```
veeenu, line 44: I'd rather not make struct members public unless there is an explicit need to do so:
```suggestion
    engine: T,
```

veeenu (changes requested):
Thank you for your contribution! Sorry for the monstruous delay but I just haven't had a lot of time to dedicate to this library lately.

The concept looks okay to me, the changes requested below are mostly stylistic. In addition to them, could you please add a `Pipeline::reset` method that just calls the same method on the engine? After that, feel free to mark this PR as non-draft and we can merge it in. Once that's done, I plan to do an immediate patch release.

Thank you!

PR #193, opened 2024-07-10
Use unicode WinApi functions everywhere to support non-ASCII inputs.

soarqin:
Tried to hook `ImmDisableIME()` out in DS3 and ER to test IME inputs for CJK languages.
Confirmed that it is working fine with these changes (was broken with all `?` characters with IME input).


veeenu:
This is great, thank you! I briefly checked out the code and it looks good. I will try and get this merged in as soon as I can.

veeenu (approved):

PR #204, opened 2024-10-04
Include the function MH_RemoveHook

vsylva:
Bump minhook to https://github.com/TsudaKageyu/minhook/commit/c1a7c3843bd1a5fe3eb779b64c0d823bca3dc339


veeenu:
Thank you for this! Please allow me a bit of time to vet this as subtle changes on that layer might have ripple side effects.

Also, the CI fail is due to new lints -- I'll have to go and take care of them off of `main` which might take a while too.

PR #210, opened 2024-10-31
Check memory protection per page

cryeprecision:
Memory protection flags are set per page so it's enough to check them at page boundaries. (See [here](https://learn.microsoft.com/en-us/windows/win32/api/memoryapi/nf-memoryapi-virtualquery#parameters))

This simplifies the code for #209 ([relevant diff](https://github.com/veeenu/hudhook/pull/209/commits/55700fdaf645d0401abe7b9f5571135697afbf76#diff-8841888df06c36a6632525a99403e18778b7de8b8efee5310396fb7b9367ef69R113-R145)) to [this](https://github.com/veeenu/hudhook/pull/210/commits/4cbabb71ad8ac35a943811d1af159839253502d3#diff-8841888df06c36a6632525a99403e18778b7de8b8efee5310396fb7b9367ef69)

I've added a helper function that checks the memory protection at each page boundary in a memory region and returns a slice containing all consecutive readable elements starting from the start of the region.

I've stared at my implementation for a bit and I *think* the logic is correct.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,6 +86,7 @@ features = [
   "Win32_System_Diagnostics_ToolHelp",
   "Win32_System_LibraryLoader",
   "Win32_System_Memory",
+  "Win32_System_SystemInformation",
   "Win32_System_SystemServices",
   "Win32_System_Threading",
   "Win32_UI_Input_KeyboardAndMouse",
--- a/src/hooks/dx12.rs
+++ b/src/hooks/dx12.rs
@@ -7,7 +7,7 @@ use std::sync::OnceLock;
 use imgui::Context;
 use once_cell::sync::OnceCell;
 use parking_lot::Mutex;
-use tracing::{error, trace};
+use tracing::{error, trace, warn};
 use windows::core::{Error, Interface, Result, HRESULT};
 use windows::Win32::Foundation::BOOL;
 use windows::Win32::Graphics::Direct3D::D3D_FEATURE_LEVEL_11_0;
@@ -24,7 +24,6 @@ use windows::Win32::Graphics::Dxgi::{
     DXGI_SWAP_CHAIN_FLAG_ALLOW_MODE_SWITCH, DXGI_SWAP_EFFECT_FLIP_DISCARD,
     DXGI_USAGE_RENDER_TARGET_OUTPUT,
 };
-use windows::Win32::System::Memory::{VirtualQuery, MEMORY_BASIC_INFORMATION, PAGE_READWRITE};
 
 use super::DummyHwnd;
 use crate::mh::MhHook;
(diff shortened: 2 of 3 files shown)


cryeprecision:
> For starters, the PR itself and the code style looks absolutely excellent

Thanks! :)

> So I wonder if my original intention was to check the protection of that page where the command queue is, actually.

Yeah that would scanning the memory pointed to by each pointer reasonable, like you said.

> Have I gotten some of this wrong?

What you wrote fits (and cleared some stuff up for me).

> I think we could get away with not doing the check altogether

I think so too and event if the pointer is at the bottom of the struct, the whole struct should be within readable memory. This should be perfectly safe as long as they don't remove the pointer and we start reading outside the struct.

The caveat you pointed out got me thinking that I might have misunderstood the code and broke something but after looking over both the old and new snippets again I'm sure it's still doing the same thing.

I have no DirectX background (which is why I'm using this library :P), so I thought the `swap_chain_ptr` was a pointer to a vtable and that you're looking through the vtable to find a pointer to the `command_queue` object (which makes no sense in hindsight because a vtable only contains pointers to functions and not to other objects but I didn't think of that earlier). Now I understand that we're actually scanning struct members for a pointer.

I found [this](https://stackoverflow.com/a/40379046) post and the offset from the answer is different from the one I get when testing (for me the command queue is at `+0x168`) so doing it this way is definitely better than hardcoding an offset.

I attached a debugger to a [DirectX 12 example](https://learn.microsoft.com/en-us/samples/microsoft/directx-graphics-samples/d3d12-hello-world-samples-win32/) and set a hardware breakpoint to see what reads/writes to `swap_chain+0x168` (i.e., sets or reads the command queue pointer) but it didn't get hit :/
The idea was to see which functions access the struct member, open one in a disassembler, and generate a signature for the assembly that accesses the struct member with the hope that different versions of the DirectX 12 library will match the signature and only the offset will be different.

veeenu:
> I think so too and event if the pointer is at the bottom of the struct, the whole struct should be within readable memory. This should be perfectly safe as long as they don't remove the pointer and we start reading outside the struct.

Yeah, though I don't feel like risking it -- for example, on Wine the command queue is the first qword immediately after the vtable (e.g. offset +1).

> The caveat you pointed out got me thinking that I might have misunderstood the code and broke something but after looking over both the old and new snippets again I'm sure it's still doing the same thing.

Yep, that was definitely me mixing things up with the attempt before.

> I found [this](https://stackoverflow.com/a/40379046) post and the offset from the answer is different from the one I get when testing (for me the command queue is at `+0x168`) so doing it this way is definitely better than hardcoding an offset.

I discarded doing outright that because I got burned by it already a few years ago, when a new d3d11.dll was released and they shuffled the pointers around. I actually wanted to go the DirectComposition way myself which would've allowed me to not maintain 4 separate renderers but when I was already done I went ahead, tested it on Linux and discovered that that's not implemented by Wine, so had to roll everything back and start over. :( 

> I attached a debugger to a [DirectX 12 example](https://learn.microsoft.com/en-us/samples/microsoft/directx-graphics-samples/d3d12-hello-world-samples-win32/) and set a hardware breakpoint to see what reads/writes to `swap_chain+0x168` (i.e., sets or reads the command queue pointer) but it didn't get hit :/ The idea was to see which functions access the struct member, open one in a disassembler, and generate a signature for the assembly that accesses the struct member with the hope that different versions of the DirectX 12 library will match the signature and only the offset will be different.

I think that would be even trickier. I expect that a change in struct layout could send the optimizer down through different paths and result in different looking assembly at least for some functions, so that's kinda wobbly to rely upon.

veeenu (approved):

PR #225, opened 2025-03-30
Make using fences on DX12 optional via cargo feature

OneXDeveloper:
I noticed pretty significant FPS drops in my games when injecting even the most basic hello world examples. I was able to track it down to this fence usage. As an example, in the game I am working on I get ~460FPS sitting in the town, but as soon as I inject the hello world example from the README it drops down to around ~240FPS. With this change I stick within ~10FPS of the game without injection. I didn't notice any weird synchronization issues on my game but I realize that this could potentially cause some if we remove it outright for all which is why I made it an optional feature for DX12 only and enabled it by default to retain existing functionality. 

P.S, your discord links in the readme/troubleshooting guide on all your repos has expired. :)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,7 +19,7 @@ targets = [
 ]
 
 [features]
-default = ["dx9", "dx11", "dx12", "opengl3", "inject"]
+default = ["dx9", "dx11", "dx12", "fenced_dx12", "opengl3", "inject"]
 dx9 = []
 dx11 = []
 dx12 = []
@@ -28,6 +28,7 @@ inject = []
 imgui-freetype = ["imgui/freetype"]
 imgui-docking = ["imgui/docking"]
 imgui-tables-api = ["imgui/tables-api"]
+fenced_dx12 = []
 
 [[example]]
 name = "simple_hook"
--- a/src/renderer/backend/dx12.rs
+++ b/src/renderer/backend/dx12.rs
@@ -15,8 +15,11 @@ use windows::Win32::Graphics::Direct3D12::*;
 use windows::Win32::Graphics::Dxgi::Common::*;
 
 use crate::renderer::RenderEngine;
-use crate::util::{self, Fence};
+use crate::util;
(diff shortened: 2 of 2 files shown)


veeenu (changes requested):
I agree that the current synchronization model in hudhook is not great, as it basically blocks the CPU while the GPU is working, but I also think this PR is not the right approach to support a large number of games and creates extra maintenance burden due to having to test and maintain the version with the feature flag and the version without.

I'm not opposed to introducing changes though as ~4ms drop (if my math is correct) is a _long_ time... Let's discuss alternative approaches.

We are injecting rendering commands just before a call to `Present`, which means that we _have_ to have completed rendering before that moment. I'm not sure what would happen if e.g. we get to the next `Present` and start rendering to the render target while the previous frame is still in flight in a case where there is no double buffering. My guess is that there would be a hard crash due to a locked resource being touched by something else.

On the other hand, we _do_ know that hudhook's code runs only in between calls to `Present` so arguably the previous frame has surely been rendered to its render target, but... what happens if hudhook is not done rendering by the time the buffer is swapped?

Another point against introducing this change is that the official imgui dx12 backend does the [same thing](https://github.com/ocornut/imgui/blob/8098d79ca2e7787b082189623c5ffbba2cca6d4d/backends/imgui_impl_dx12.cpp#L411-L434) which makes sense as the entire context is thread local and you can't rely on the host doing their synchronization in a compatible way. I suspect that when you are trying to provide a generic abstracted solution, like we are, there is some cost to be paid.

All in all, if we can find a more elegant solution that saves time without requiring a feature flag I'd be up for it, but I'd like to have soundness for the concurrency part -- could be e.g. waiting on the previous fence at the start of the frame and increasing it at the end, or maybe using multiple fences so that you only wait on the one for the current frame in the case of multiple buffering.

(I want to try running this against a few dx12 games anyway just to see what happens... though I no longer have Windows so that'll really be running against Vulkan)

Issue #214, opened 2024-11-07
Why MSRV is not specified


luadebug:
https://github.com/veeenu/hudhook/blob/465cb118f3fee0b7d57f83f19a53bb846f9527b3/Cargo.toml#L11
I think it should be already specified & exist there?
```toml
rust-version = "1.80.1"
```
```pwsh
E:\hudhook git:[main]
cargo msrv find
  [Meta]   cargo-msrv 0.16.1                                                                                                                             
```
```pwsh
Compatibility Check #1: Rust 1.69.0
  [FAIL]   Is incompatible
Compatibility Check #2: Rust 1.76.0                                                                                                                      
  [FAIL]   Is incompatible
Compatibility Check #3: Rust 1.79.0                                                                                                                      
  [FAIL]   Is incompatible
Compatibility Check #4: Rust 1.81.0                                                                                                                      
  [OK]     Is compatible
Compatibility Check #5: Rust 1.80.1
  [OK]     Is compatible
Result:
   Considered (min … max):   Rust 1.56.1 … Rust 1.82.0 
   Search method:            bisect
   MSRV:                     1.80.1
   Target:                   x86_64-pc-windows-msvc
```

<details>
<summary>Compatibility Check #2: Rust 1.76.0</summary>

```pwsh
  │    Compiling windows_x86_64_msvc v0.52.6                                                                  │
  │    Compiling shlex v1.3.0                                                                                 │
  │     Checking cfg-if v1.0.0                                                                                │
  │    Compiling khronos_api v3.1.0                                                                           │
  │    Compiling autocfg v1.4.0                                                                               │
  │    Compiling parking_lot_core v0.9.10                                                                     │
  │     Checking once_cell v1.20.2                                                                            │
  │    Compiling log v0.4.22                                                                                  │
  │     Checking scopeguard v1.2.0                                                                            │
  │    Compiling xml-rs v0.8.23                                                                               │
  │     Checking regex-syntax v0.6.29                                                                         │
  │     Checking regex-syntax v0.8.5                                                                          │
  │     Checking smallvec v1.13.2                                                                             │
  │     Checking mint v0.5.9                                                                                  │
  │     Checking chlorine v1.0.12                                                                             │
  │     Checking pin-project-lite v0.2.15                                                                     │
  │     Checking bitflags v1.3.2                                                                              │
  │     Checking bitflags v2.6.0                                                                              │
  │    Compiling cc v1.1.36                                                                                   │
[...]

Issue #199, opened 2024-07-26
Switch to std `LazyLock` / `LazyCell`


veeenu:
The above mentioned features were recently stabilized, so it's time to remove the dependency on `once_cell` and bump MSRV to 1.80.

veeenu:
Wanted to do this now, but we need `once_cell_try` to land in stable before that.

PR #227, opened 2025-06-16
fix: missing size_of import

jacobtread:
Not sure how you managed to compile and release without this? The project was not building (The latest release on crates.io was not building for me either however v0.5 was building) this import was missing from the code. 

Edit: Ah I see this might have been added into the prelude in newer versions of Rust. I'm currently pinned on 1.75.0 for some other requirements, this is the only thing seeming to block compiling for that rust version though so couldn't hurt to merge since its just an added import

### DIFF
--- a/src/util.rs
+++ b/src/util.rs
@@ -3,7 +3,7 @@
 
 use std::ffi::{c_void, OsString};
 use std::fmt::Display;
-use std::mem::ManuallyDrop;
+use std::mem::{size_of, ManuallyDrop};
 use std::os::windows::ffi::OsStringExt;
 use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


veeenu:

`hudhook` has never specified MSRV because it started as a nightly-only crate, requiring some unstable APIs. We managed to drop the nightly requirement as all the APIs we needed got stabilized at some point. 

I'd be very surprised if 1.75 could get `hudhook` to compile at all, though: we use some APIs that were stabilized way after 1.75.

Out of caution, I'd only ever support [current Rust](https://github.com/veeenu/hudhook/issues/214#issuecomment-2463462681) explicitly, but possibly 1.80 would be a [viable cutoff](https://github.com/veeenu/hudhook/issues/199).

PR #231, opened 2025-08-03
Improve `eject()` crashes on DX12

zed0:
This PR reduces eject crashes by restricting the Minhook disable/uninitialize to the end of the Present call. This prevents it happening at a time when Present is wanting to use the resources that are being unhooked.

I'm fully expecting to make a lot of changes to this code, but thought I'd make a draft PR in case you have any advice on whether this is a reasonable approach.
I'm not very happy with the way synchronisation between the unhook thread and the present thread is done, but it was easy for a proof of concept.

This changes the number of crashes I get on eject while playing Shadow of the Tomb Raider with DX12 from ~30% of the time to 0% (so far).

### DIFF
--- a/src/hooks/dx12.rs
+++ b/src/hooks/dx12.rs
@@ -2,10 +2,8 @@
 
 use std::ffi::c_void;
 use std::mem;
-use std::sync::OnceLock;
 use std::sync::atomic::Ordering;
-use std::thread;
-use std::time::Duration;
+use std::sync::OnceLock;
 
 use imgui::Context;
 use once_cell::sync::OnceCell;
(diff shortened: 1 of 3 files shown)


zed0:
The code has been updated to replace busy loops with the EJECT_REQUESTED approach as suggested. This runs most of the ejection code synchronously during Present. The DLL ejection then happens in a separate thread which waits for none of the hooked functions to be running before ejecting the DLL.
The code is certainly a lot cleaner without the loops 😅

artosimonyan:
Hey @veeenu, think this will solve the issues mentioned in #139, which I'm also experiencing? If so can we get it reviewed and merged if possible?

veeenu:
Hi,

I haven't forgotten this, I've just been impossibly busy and couldn't really devote any time at all to open source lately.
I'll try to carve some time in the coming week.

veeenu (approved):
LGTM, thank you so much for this :raised_hands: 

Some of the lint errors are newer ones that come from the library itself and not this commit, so I fixed those and took the chance to bump the version number so we can issue a new release.

Feel free to rebase and address the few ones remaining, then we can merge in!

veeenu (approved):
LGTM, thank you!

PR #237, opened 2025-11-12
WndProc message filters for imgui input captures

soarqin:
We should not dispatch mouse/keyboard events to main game/application if these fields are true, as described in imgui.h comments:
```
bool        WantCaptureMouse;                   // Set when Dear ImGui will use mouse inputs, in this case do not dispatch them to your main game/application (either way, always pass on mouse inputs to imgui). (e.g. unclicked mouse is hovering over an imgui window, widget is active, mouse was clicked over an imgui window, etc.).
bool        WantCaptureKeyboard;                // Set when Dear ImGui will use keyboard inputs, in this case do not dispatch them to your main game/application (either way, always pass keyboard inputs to imgui). (e.g. InputText active, or an imgui window is focused and navigation is enabled, etc.).
```
This prevents some unintentional inputs to main game while operating on imgui windows.

### DIFF
--- a/src/renderer/msg_filter.rs
+++ b/src/renderer/msg_filter.rs
@@ -75,6 +75,11 @@ impl MessageFilter {
         match message_id {
             WM_KEYFIRST..=WM_KEYLAST => self.contains(Self::InputKeyboard),
             WM_MOUSEFIRST..=WM_MOUSELAST => self.contains(Self::InputMouse),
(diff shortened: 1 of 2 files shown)


veeenu:
If you mean to use this in the context of From Software games, unfortunately filtering the messages from the window procedure won't work. I have tried over and over. Even raw input passes through. They have some other mechanism of listening to inputs that bypasses window procedure completely but I never figured out what that is.

PR #252, opened 2026-04-14
fix: resolve DX11 fullscreen errors and washed-out colors

dontucare:
## Summary

Fixes multiple issues that occur when the hooked application switches to fullscreen mode on DirectX 11.

### Problems

1. **`Insufficient display size: 0x0` error spam** - when transitioning to fullscreen, `WM_SIZE` delivers zero dimensions, and `pipeline.resize()` blindly sets `display_size = [0, 0]`, causing an error on every frame.

2. **`Render error: HRESULT(0xFFFFFFFF)`** - the zero display size check returned a hard error instead of skipping the frame gracefully.

3. **ImGui appears washed-out/gray in fullscreen** - caused by:
   - Incorrect alpha blend factors (`SrcBlendAlpha=INV_SRC_ALPHA`, `DestBlendAlpha=ZERO` instead of `ONE`/`ZERO`), leaving the back buffer alpha < 1.0, which DXGI/DWM interprets as semi-transparency in fullscreen compositing.
   - Double gamma correction when the back buffer has an `_SRGB` format and `CreateRenderTargetView` is called with `None` desc (auto-detect inherits the `_SRGB` view, causing the GPU to apply linear to sRGB conversion on top of already-sRGB data).

4. **No `ResizeBuffers` hook for DX11** - unlike DX9 (hooks `Reset`) and DX12 (hooks `ResizeBuffers`), the DX11 hook only intercepted `Present`. When the swap chain resizes buffers during fullscreen transitions, the pipeline has no way to react.

5. **`display_size` never updated from swap chain** - the display size was only set from `GetClientRect` at init and `WM_SIZE` on resize, but never from the actual swap chain back buffer dimensions. In fullscreen, `GetClientRect` and `WM_SIZE` can report incorrect values.

6. **`display_framebuffer_scale` never updated** - on HiDPI displays, the framebuffer scale may change when switching to fullscreen, but it was always left at the default `[1.0, 1.0]`.

7. **`framebuffer_scale` not applied to viewport** - DX11 and DX12 viewports used only `display_size`, ignoring `framebuffer_scale`, causing incorrect rendering on HiDPI displays.

### Changes

- **`pipeline.resize()`**: skip updates when width/height are zero to preserve previous display size
- **`pipeline.render()`**: return `Ok(())` (skip frame with a warning) instead of `Err` when display size is insufficient
- **`Pipeline::update_display_size_from_swap_chain()`**: new method that updates both `display_size` and `display_framebuffer_scale` (from window DPI via `GetDpiForWindow`)
- **DX11 hooks**: add `IDXGISwapChain::ResizeBuffers` hook; update display size from `swap_chain.GetDesc()` before each frame and after `ResizeBuffers`
- **DX11/DX12 blend state**: `SrcBlendAlpha=ONE`, `DestBlendAlpha=ZERO` - ensures back buffer alpha stays 1.0 (matches reference `imgui_impl_dx11.cpp`)
- **DX11/DX12 viewport**: apply `framebuffer_scale` to viewport dimensions
- **DX11 RTV creation**: derive format and `ViewDimension` from back buffer `GetDesc()`, strip `_SRGB` suffix to prevent double gamma; fall back to auto-detect (`None`) if explicit desc fails (handles MSAA, array textures, etc.)
- **Cargo.toml**: add `Win32_UI_HiDpi` feature for `GetDpiForWindow`

## Testing

- All changes compile cleanly with `cargo check` and `cargo check --tests`
- Tested in a DX11 application: fullscreen transition no longer produces error spam, ImGui colors remain correct

### DIFF
--- a/src/renderer/backend/dx11.rs
+++ b/src/renderer/backend/dx11.rs
@@ -82,18 +82,26 @@ impl RenderEngine for D3D11RenderEngine {
         unsafe {
             let state_backup = StateBackup::backup(&self.device_context);
 
+            let rtv_desc = {
+                let mut tex_desc = D3D11_TEXTURE2D_DESC::default();
+                render_target.GetDesc(&mut tex_desc);
+                let format = match tex_desc.Format {
+                    DXGI_FORMAT_R8G8B8A8_UNORM_SRGB => DXGI_FORMAT_R8G8B8A8_UNORM,
+                    DXGI_FORMAT_B8G8R8A8_UNORM_SRGB => DXGI_FORMAT_B8G8R8A8_UNORM,
+                    DXGI_FORMAT_B8G8R8X8_UNORM_SRGB => DXGI_FORMAT_B8G8R8X8_UNORM,
+                    other => other,
+                };
+                D3D11_RENDER_TARGET_VIEW_DESC {
+                    Format: format,
+                    ViewDimension: D3D11_RTV_DIMENSION_TEXTURE2D,
+                    Anonymous: D3D11_RENDER_TARGET_VIEW_DESC_0 {
+                        Texture2D: D3D11_TEX2D_RTV { MipSlice: 0 },
+                    },
+                }
+            };
+
             let render_target: ID3D11RenderTargetView = util::try_out_ptr(|v| {
-                self.device.CreateRenderTargetView(
-                    &render_target,
-                    Some(&D3D11_RENDER_TARGET_VIEW_DESC {
-                        Format: DXGI_FORMAT_B8G8R8A8_UNORM,
-                        ViewDimension: D3D11_RTV_DIMENSION_TEXTURE2D,
-                        Anonymous: D3D11_RENDER_TARGET_VIEW_DESC_0 {
-                            Texture2D: D3D11_TEX2D_RTV { MipSlice: 0 },
-                        },
-                    }),
-                    Some(v),
-                )
+                self.device.CreateRenderTargetView(&render_target, Some(&rtv_desc), Some(v))
             })?;
 
             self.device_context.OMSetRenderTargets(Some(&[Some(render_target)]), None);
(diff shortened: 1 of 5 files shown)


veeenu:

Please allow me a bit of time to test this against [my](https://github.com/veeenu/eldenring-practice-tool) [usual](https://github.com/veeenu/darksoulsiii-practice-tool) [harnesses](https://github.com/veeenu/sekiro-practice-tool) and do a proper code review but I like this already. :raised_hands:

src/renderer/backend/dx11.rs
```diff
@@ -82,9 +82,54 @@ impl RenderEngine for D3D11RenderEngine {
...
+            };
+
+            let render_target: ID3D11RenderTargetView = match util::try_out_ptr(|v| {
+                self.device.CreateRenderTargetView(&render_target, Some(&rtv_desc), Some(v))
+            }) {
+                Ok(rtv) => rtv,
+                Err(_) => util::try_out_ptr(|v| {
+                    self.device.CreateRenderTargetView(&render_target, None, Some(v))
+                })?,
+            };
```
veeenu: Nit: this would probably look more concise as something like

```rust
util::try_out_ptr(...)
  .or_else(|_| util::try_out_ptr(...))?
```

veeenu (commented):
This looks nice, thank you. Since I no longer have Windows, could you please provide before-and-after screenshots of the double SRGB and alpha blend issues just for the sake of documenting them here?

I have a small ask and then if you could please run [nightly cargo fmt and clippy](https://github.com/veeenu/hudhook/blob/main/CONTRIBUTING.md#preparing-a-pull-request) to get CI to pass we can merge in.

dontucare:
**before:**
Screenshot (2)

Screenshot (4)

**after:**
Screenshot (1)

Screenshot (3)

veeenu (approved):
Looks great, thank you! 🙌

Issue #256, opened 2026-06-12
DX12 hook fails in Elden Ring 1.16.1: command queue pointer not found in swap chain


chozandrias76:
## Summary

`hudhook` 0.9.1 fails to initialize the DX12 renderer in Elden Ring 1.16.1 because `InitializationContext::check_command_queue()` never finds the observed `ID3D12CommandQueue` pointer inside the observed `IDXGISwapChain3` object memory.

The DLL injects and the game remains stable, but the ImGui overlay never appears.

## Environment

- Game: Elden Ring 1.16.1
- Renderer: DX12
- hudhook: 0.9.1 / `6a361de92396bcbb1f46cec31348fd73344ed031`
- Usage: Rust DLL loaded into the game process, using `hudhook::hooks::dx12::ImguiDx12Hooks`
- Tested with other mods disabled; only this DLL loaded via LazyLoader

## Observed logs

The log repeats every frame:

```text
hudhook::hooks::dx12: Couldn't find command queue pointer in swap chain struct (512 out of 512 pointers were readable)
hudhook::hooks::dx12: Initialization context incomplete
hudhook::hooks::dx12: Render error: HRESULT(0xFFFFFFFF)
```

## Expected behavior

The DX12 initialization should accept the command queue associated with the presented swap chain and initialize the render context, allowing the overlay to render.

## Actual behavior

`Present` and `ExecuteCommandLists` appear to be hooked, but the command queue/swapchain association check never succeeds. Because `check_command_queue()` depends on finding the raw command queue pointer by scanning private swapchain object memory, initialization remains incomplete indefinitely.

## Notes / related upstream context

This looks related to the same heuristic discussed in:

- #209, which introduced matching observed command queues against memory near the swap chain
- #226, where another DX12 game reportedly worked around a hook failure by making `check_command_queue()` return `true`

This issue is specifically for Elden Ring 1.16.1 and the current 0.9.1 DX12 path. It seems possible that the DXGI/driver/game/overlay stack no longer stores the command queue pointer in the swapchain object in a scan-friendly way, or that wrapper/proxy COM objects make pointer identity differ from the scanned memory.

## Possible fix direction

A robust fix may need to avoid requiring private swapchain-memory layout to prove the queue association. For example:

- allow an opt-in relaxed mode that accepts the first observed command queue after a swapchain is known,
- hook swapchain creation to record the queue passed to `CreateSwapChainForHwnd`, or
[...]

PR #257, opened 2026-06-12
Add opt-in DX12 command queue device fallback

chozandrias76:
## Summary

Adds an opt-in DirectX 12 initialization fallback behind the `dx12-fallback-device-check` feature.

When the existing swap-chain memory scan cannot find the observed `ID3D12CommandQueue` pointer, the fallback accepts the command queue if it belongs to the same `ID3D12Device` as the swap chain.

This keeps current default behavior unchanged while providing an escape hatch for games/drivers where the private DXGI swap-chain object layout does not expose the command queue pointer in a scan-friendly way.

## Motivation

Fixes/addresses #256.

In Elden Ring 1.16.1 under Proton, the current DX12 hook repeatedly failed with:

```text
Couldn't find command queue pointer in swap chain struct (512 out of 512 pointers were readable)
Initialization context incomplete
Render error: HRESULT(0xFFFFFFFF)
```

The DLL loaded and the game remained stable, but the ImGui overlay never rendered because initialization never reached `Complete`.

## Validation

Checked both the default path and the new opt-in feature:

```bash
cargo xwin check --target x86_64-pc-windows-msvc --no-default-features --features dx12,imgui-tables-api
cargo xwin check --target x86_64-pc-windows-msvc --no-default-features --features dx12,imgui-tables-api,dx12-fallback-device-check
```

Also tested in-game with a downstream Elden Ring DLL using the feature. The overlay rendered successfully on the Elden Ring main menu. The log no longer showed the repeated command-queue/init spam; only one early startup `Initialization context incomplete` appeared before the render context completed.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,11 @@ default = ["dx9", "dx11", "dx12", "opengl3", "inject"]
 dx9 = []
 dx11 = []
 dx12 = []
+# Allow the DirectX 12 hook to accept a command queue from the same D3D12
+# device when the private swap-chain memory scan cannot find the queue pointer.
+# This is useful for games/drivers whose DXGI swap-chain implementation does
+# not expose the command queue in a scan-friendly layout.
+dx12-fallback-device-check = ["dx12"]
 opengl3 = ["dep:gl_generator"]
 inject = []
 imgui-freetype = ["imgui/freetype"]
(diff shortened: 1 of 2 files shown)


veeenu:
Thank you for your contribution and sorry for the late response -- I don't have a lot of time for maintenance lately but hopefully I should in the coming weeks. Could you please try the branch in #258 and see if it fixes the problem for you instead? In that case, I'd rather merge that one since it's more comprehensive.

veeenu:
Superseded by #258. Thank you for taking the time for your contribution :raised_hands:

PR #258, opened 2026-07-02
fix(dx12): stabilize swap-chain resize handling

soarqin:
* Capture DX12 command queues from swap-chain creation and ResizeBuffers1 paths, validate active swap-chain/device/queue identity, and rebuild the overlay pipeline safely after resize, format, buffer-count, or device changes.
* Thread RTV format and NodeMask through the DX12 renderer, wait for overlay fences before resize, fix texture upload fence ordering, and extend the DX12 harness to exercise real ResizeBuffers.

This PR should also fix the issue noted in #257.

It is ported from recent changes for my EROverlay project which are inspired by RadialMenu to support MSHybrid GPU switching mechanism and resizing of rendering context.

### DIFF
--- a/src/hooks/dx12.rs
+++ b/src/hooks/dx12.rs
@@ -9,7 +9,8 @@ use imgui::Context;
 use once_cell::sync::OnceCell;
 use parking_lot::Mutex;
 use tracing::{debug, error, trace, warn};
-use windows::core::{Error, Interface, Result, BOOL, HRESULT};
+use windows::core::{Error, IUnknown, Interface, Result, BOOL, HRESULT};
+use windows::Win32::Foundation::{HWND, LUID};
 use windows::Win32::Graphics::Direct3D::D3D_FEATURE_LEVEL_11_0;
 use windows::Win32::Graphics::Direct3D12::{
     D3D12CreateDevice, ID3D12CommandList, ID3D12CommandQueue, ID3D12Device, ID3D12Resource,
@@ -20,9 +21,11 @@ use windows::Win32::Graphics::Dxgi::Common::{
     DXGI_MODE_SCANLINE_ORDER_UNSPECIFIED, DXGI_RATIONAL, DXGI_SAMPLE_DESC,
 };
 use windows::Win32::Graphics::Dxgi::{
-    CreateDXGIFactory2, IDXGIFactory2, IDXGISwapChain, IDXGISwapChain3, DXGI_CREATE_FACTORY_FLAGS,
-    DXGI_SWAP_CHAIN_DESC, DXGI_SWAP_CHAIN_FLAG_ALLOW_MODE_SWITCH, DXGI_SWAP_EFFECT_FLIP_DISCARD,
-    DXGI_USAGE_RENDER_TARGET_OUTPUT,
+    CreateDXGIFactory2, IDXGIFactory, IDXGIFactory2, IDXGIOutput, IDXGISwapChain, IDXGISwapChain1,
+    IDXGISwapChain2, IDXGISwapChain3, DXGI_CREATE_FACTORY_FLAGS, DXGI_ERROR_DEVICE_REMOVED,
+    DXGI_ERROR_DEVICE_RESET, DXGI_PRESENT_PARAMETERS, DXGI_SWAP_CHAIN_DESC, DXGI_SWAP_CHAIN_DESC1,
+    DXGI_SWAP_CHAIN_FLAG_ALLOW_MODE_SWITCH, DXGI_SWAP_CHAIN_FULLSCREEN_DESC,
+    DXGI_SWAP_EFFECT_FLIP_DISCARD, DXGI_USAGE_RENDER_TARGET_OUTPUT,
 };
 
 use super::DummyHwnd;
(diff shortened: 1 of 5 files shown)


src/hooks/dx12.rs
```diff
@@ -170,17 +230,345 @@ impl InitState {
...
+    command_queue_device_identity: usize,
+    adapter_luid: LUID,
+    rtv_format: DXGI_FORMAT,
+    buffer_count: u32,
+    hwnd: usize,
+}
+
+fn format_luid(luid: LUID) -> String {
+    format!("{:08x}:{:08x}", luid.HighPart, luid.LowPart)
+}
```
veeenu: I think it would be a better idea to have a `std::fmt::Display` implementation over a wrapper object here so we can avoid allocating. Something like:

```rust
struct FmtLuid(LUID);

impl Display for FmtLuid {
    ...
}
```

This way one could do:

```rust
println!("Some message related to luid: {}", FmtLuid(luid));
```

src/hooks/dx12.rs
```diff
@@ -257,8 +732,149 @@ unsafe extern "system" fn dxgi_swap_chain_resize_buffers_impl(
...
+                {
+                    warn!("Could not update DX12 display size after ResizeBuffers: {e:?}");
+                }
+            } else {
+                warn!("Could not lock DX12 pipeline to update display size after ResizeBuffers");
+            }
+        }
+    }
+
+    result
```
veeenu, line 780: I'd like to reduce the nesting a little bit with early returns here:

- Considering that we return `result` as is here just after the `if`, we could just add a `?` at the top assignment and return `Ok(result)` at the bottom instead. Alternatively, invert `if result.is_ok() { big_branch() }` to `if result.is_err() { return result; }`
- We could furthermore do an early return on the first if let:
  ```rust
  let (Some(swap_chain3), ...) = (...) else { return Ok(result) };
  ```
- We could use let-else and if-let for the rest:
  ```rust
  let Some(mut pipeline) = pipeline.try_lock() else { warn!("..."); return Ok(result) };
  if let Err(e) = ... // as-is
  ```

src/hooks/dx12.rs
```diff
@@ -257,8 +732,149 @@ unsafe extern "system" fn dxgi_swap_chain_resize_buffers_impl(
...
+                } else {
+                    warn!(
+                        "Could not lock DX12 pipeline to update display size after SetSourceSize"
+                    );
+                }
+            }
+        }
+    }
+
+    result
```
veeenu, line 805: Similar idea to the above comment, we should reduce nesting where possible.

src/hooks/dx12.rs
```diff
@@ -257,8 +732,149 @@ unsafe extern "system" fn dxgi_swap_chain_resize_buffers_impl(
...
+                            "Could not lock DX12 pipeline to update display size after \
+                             ResizeBuffers1"
+                        );
+                    }
+                }
+            },
+        }
+    }
+
+    result
```
veeenu, line 870: Same as above.

veeenu (changes requested):
Thank you so much for your contribution 🙌 this looks like a big improvement over what we have. I'm just a bit worried of hooking so many methods but maybe it is necessary. Do you have a ballpark estimate of what the additional overhead might be and how many of those hooks are on the hot path?

I have a handful of style-related comments only for now, haven't been able to dive deep into the code and there's some concepts I'm not super familiar with. Besides I haven't had a chance to test it on my machines yet, and won't be able to do so at all on Windows itself, so I'll trust you if you say this works on Windows. Have you had the chance to test it with earlier versions of Elden Ring? I use that as a decent proxy for "the changes are not breaking existing functionality".

Thank you again!

src/hooks/dx12.rs
soarqin: resolved in 26efcc2

src/hooks/dx12.rs
soarqin: resolved in 26efcc2

src/hooks/dx12.rs
soarqin: resolved in 26efcc2

src/hooks/dx12.rs
soarqin: resolved in 26efcc2

soarqin:
> Thank you so much for your contribution 🙌 this looks like a big improvement over what we have. I'm just a bit worried of hooking so many methods but maybe it is necessary. Do you have a ballpark estimate of what the additional overhead might be and how many of those hooks are on the hot path?

Part of them are only used to handle dx12 context recreation after gpu-switching / resizing / driver lost, to make imgui context handle/width/height synced.
No need to worry about over-hooking coz they are not used in any other place.

> I have a handful of style-related comments only for now, haven't been able to dive deep into the code and there's some concepts I'm not super familiar with. Besides I haven't had a chance to test it on my machines yet, and won't be able to do so at all on Windows itself, so I'll trust you if you say this works on Windows. Have you had the chance to test it with earlier versions of Elden Ring? I use that as a decent proxy for "the changes are not breaking existing functionality".

I have tested it on 1.02, 1.07, 1.10 and 1.16 and all working fine.

> Thank you again!

I may do a similar work for dx11 later, to make ds3/sekiro practice tool take benefit too.

veeenu:
LGTM, thank you so much! :raised_hands:

veeenu (approved):

PR #260, opened 2026-07-30
Add DirectX9Ex support

super-continent:
This adds support for hooking DirectX9Ex, which is used in a few different programs that I want to work on using the crate (great work btw, thanks for maintaining something like this!) The code itself is mostly a copy of the current D3D9 code, with all the modifications needed to support the `Ex` variants of the hooked functions

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,7 @@ name = "hudhook"
 version = "0.9.2"
 edition = "2021"
 rust-version = "1.85"
-description = "A graphics API hook with dear imgui render loop. Supports DirectX 9, 11, 12, and OpenGL 3."
+description = "A graphics API hook with dear imgui render loop. Supports DirectX 9, 9Ex, 11, 12, and OpenGL 3."
 homepage = "https://github.com/veeenu/hudhook"
 repository = "https://github.com/veeenu/hudhook"
 documentation = "https://veeenu.github.io/hudhook"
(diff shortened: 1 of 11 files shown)


veeenu:
Thank you for your contribution! :raised_hands: 

Haven't looked it over in depth yet, but I wonder if there is a lot of code that's copied we might make this PR smaller by sharing common behavior rather than duplicating functions. This would help a lot with maintainability.

super-continent:
This might be possible, the Present and Reset functions are the same as DX9 in this case. I'm unsure how to cleanly structure a merge like that though. An impl macro could work but that could end up being hard to read depending on how complex we need to make it

veeenu:
Hi, sorry this took a while but I haven't really been able to spend any time working on open source lately :sweat: 

Apparently `IDirect3D9Ex` implements `Deref<Target=IDirect3D9>`. Probably at least `init_pipeline`, `render` and `reset_pipeline` could be removed and called something like

```rust
use super::dx9::{render, init_pipeline, reset_pipeline);

  // later...
  if let Err(e) = render(&device.deref()) {
      error!("Render error: {e:?}");
  }
```

Could structure the module like

```
src/hooks/dx9/mod.rs
src/hooks/dx9/ex.rs
```

and remove those functions and `PIPELINE` and `RENDER_LOOP from your module. I'm sure there are other avenues I haven't considered yet.

Possibly also the harness could be deduplicated, but I'm not really worried about that. It's fine for it to stay duplicated.

super-continent:
This should be a more maintainable version of the change.

